PR #7, opened 2025-09-01
chore: track .lock files, update .gitignore

yuzurihaaa:
Hi. Thanks for the repo!

Playing around with this repo and realised that these part could be improved in this repo.

1. Track `pubspec.lock` so that other contributors will use the same package version when fetching dependencies.
2. Ignore files according to https://github.com/flutter/flutter/blob/master/packages/flutter_tools/templates/package/.gitignore.tmpl


yuzurihaaa:
Let me know if there's files that I should keep tracking. Currently I remove those files according to new .gitignore change.

michaelbushe (commented):
Yes, .dart_tool should be ignored and pubspec.lock should be ignored.  I prefer .idea is checked in since it has IDE runtime configs.

PR #11, opened 2025-10-20
fix: Add platform-specific EnvironmentService implementations for web support

YumNumm:
## Overview
This PR fixes the environment variable access issue on web platforms by introducing platform-specific implementations of `EnvironmentService`.

## Problem
The original `EnvironmentService` implementation relied on `dart:io`'s `Platform.environment`, which is not available on web platforms. This caused runtime errors when trying to access environment variables in web builds.

image

## Solution
Refactored `EnvironmentService` to use conditional exports with platform-specific implementations:

- **`environment_service.dart`**: Defines the `EnvironmentServiceInterface` and exports the appropriate implementation based on the platform
- **`environment_service_io.dart`**: IO platform implementation using both `Platform.environment` and `String.fromEnvironment`
- **`environment_service_web.dart`**: Web platform implementation using only `String.fromEnvironment` (dart:io unavailable)

### DIFF
--- a/lib/src/environment/environment_service/environment_service.dart
+++ b/lib/src/environment/environment_service/environment_service.dart
@@ -2,7 +2,7 @@
 // Copyright 2025, Michael Bushe, All rights reserved.
 
 export 'environment_service_io.dart'
-    if (dart.library.js_interop) 'environment_service_web.dart' ;
+    if (dart.library.js_interop) 'environment_service_web.dart';
 
 /// Interface for accessing environment variables in a consistent manner.
 ///


lib/src/environment/environment_service/environment_service_web.dart
```diff
@@ -29,11 +29,12 @@ class EnvironmentService {
...
   /// @param key The name of the environment variable to retrieve
   /// @return The value of the environment variable, or null if not found
+  @override
   String? getValue(String key) {
     if (_useTestEnvironment) {
       return _testEnvironment[key];
     }
-    return io.Platform.environment[key];
+    return String.fromEnvironment(key);
   }
```
YumNumm: memo-badge
On web platform, you can retrieve values defined with `dart-define` by executing `String.fromEnvironment`.

lib/src/environment/otel_env.dart
```diff
@@ -2,7 +2,7 @@
 // Copyright 2025, Michael Bushe, All rights reserved.
 
 import 'package:dartastic_opentelemetry_api/dartastic_opentelemetry_api.dart';
-import 'environment_service.dart';
+import 'environment_service/environment_service.dart';
```
YumNumm: memo-badge
fixes export path

PR #28, opened 2026-04-08
fix: SimpleLogRecordProcessor.shutdown() skips forceFlush

robert-northmind:
## Summary
`SimpleLogRecordProcessor.shutdown()` sets `_isShutdown = true` before
calling `forceFlush()`, but `forceFlush()` returns early when
`_isShutdown` is true — making the flush a no-op during shutdown.

The OTel spec requires that shutdown MUST include the effects of
ForceFlush. `BatchLogRecordProcessor` already handles this correctly
by flushing before setting the flag.

## Changes
- Reorder `shutdown()` to call `forceFlush()` before setting `_isShutdown`
- Add regression test verifying flush is called before shutdown
- Update CHANGELOG

## Spec reference
https://opentelemetry.io/docs/specs/otel/logs/sdk/#logrecordprocessor

> Shutdown SHOULD provide a way to let the caller know whether it
> succeeded, failed or timed out. Shutdown MUST include the effects
> of ForceFlush.

## Test plan
- [x] `./tool/test.sh` — all tests pass (1 pre-existing failure unrelated to this change)
- [x] `dart analyze` — no issues
- [x] `dart format` — no changes needed
- [x] New regression test verifies `forceFlush` is called before `shutdown` on the exporter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Fixed
+- Fixed SimpleLogRecordProcessor.shutdown() not flushing pending exports
+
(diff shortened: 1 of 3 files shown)


michaelbushe (approved):
Good catch, thank you.

PR #37, opened 2026-05-10
fix: shut down default LoggerProvider in OTel.shutdown (#33)

### LINKED ISSUE #33: OTel.shutdown() hangs in short-lived Dart CLI binaries (BatchSpanProcessor / OTLP gRPC exporter)
## Summary

Reproducible from the dart-otel-reference-demo's swarm runner: a short-lived Dart CLI that initializes OpenTelemetry, makes one request, then calls `OTel.shutdown()` does not exit. Process stays alive indefinitely (observed >30 minutes, identical PIDs from start) with the main thread parked in `_pthread_cond_wait` inside `Dart_RunLoop`.

The demo's swarm runner spawns 200 of these CLI invocations; all 10 parallel children hang on the FIRST batch and the script never finishes. Driving requests with curl (no in-process OTel SDK) works fine, which isolates the hang to the SDK shutdown path.

## Repro

```dart
import 'package:dartastic_opentelemetry/dartastic_opentelemetry.dart';

Future<void> main() async {
  await OTel.initialize(
    serviceName: 'repro',
    serviceVersion: '0.0.0',
    // OTLP gRPC default endpoint; can also be reproduced when the
    // endpoint is reachable, so it isn't strictly an unreachable-
    // backend issue.
  );
  final tracer = OTel.tracerProvider().getTracer('repro');
  final span = tracer.startSpan('one', kind: SpanKind.internal);
  span.end();

  await OTel.tracerProvider().forceFlush();
  await OTel.shutdown();          // <-- never returns
  print('past shutdown');         // never printed
}
```

`dart run repro.dart` against `dartastic_opentelemetry: 1.1.0-beta` does not exit. `kill -9` is the only way out. macOS arm64, Dart SDK 3.11.5.

`sample <pid>` output of a hung process — main thread:

```
Thread_xxx  DispatchQueue_1: com.apple.main-thread
  start  (in dyld) + 6076
    ... dartvm ...
      Dart_RunLoop  (in dartvm) + 256
        ... dartvm ...
          _pthread_cond_wait  (in libsystem_pthread.dylib) + 984
            __psynch_cvwait  (in libsystem_kernel.dylib) + 8
```

Plus a `dart:io EventHandler` thread parked on `kevent` and one or more `DartWorker` threads also in `_pthread_cond_wait`. No active work; the runtime is just waiting for something that never finishes.

## Hypothesis

Likely candidates for what `shutdown()` is waiting on:

1. **`BatchSpanProcessor`'s drain queue** — if the queue isn't drained or the export future never resolves (e.g. an in-flight gRPC call that's stuck), `shutdown()` blocks forever waiting.
2. **OTLP gRPC channel cleanup** — gRPC for Dart uses long-lived HTTP/2 connections; shutting them down on a connection that never completed a handshake (or completed but is mid-export) may leave a future unresolved.
3. **`ConsoleExporter` (in `CompositeMetricExporter` default)** — less likely, but the metric reader's drain in shutdown might also be involved.

Repros against the OTLP gRPC exporter; haven't tested with `secure: false` + a working endpoint vs. the default `secure: true` against an unreachable one. The hang reproduces with both the demo's local stack (lgtm container reachable) and without it (no backend). The shutdown path should tolerate both.

## What this affects

- Any short-lived Dart CLI/binary that uses the SDK and exits on its own (no long-running process where SIGTERM is the natural shutdown trigger). Scripts. CLI tools. Tests that initialize the real SDK and don't `dart:io exit()` to bypass cleanup.
- The demo's `load/run_swarm.sh` — current workaround is to drive load with curl instead of the CLI, which loses one of the demo's better stories (CLI as the trace root span).

## What works around it

- Forcibly `exit(0)` from the CLI after a short post-shutdown timeout. Workable but loses the SDK's structured drain.
- Skip `OTel.shutdown()` entirely; let the OS reclaim. Loses any in-flight spans the BatchSpanProcessor hadn't yet exported — exactly what shutdown is supposed to prevent.

Neither is a fix for the actual demo. The demo is a reference implementation people will copy-paste; whatever the right shutdown pattern is, this needs to be it, and `OTel.shutdown()` needs to terminate.

Happy to attach a minimal repro repo if useful.

michaelbushe:
Closes #33.

## Summary

`OTel.shutdown()` was iterating over tracer providers and meter providers but never calling `loggerProvider.shutdown()`. With logs enabled by default (`enableLogs: true`), the default `LoggerProvider` instantiates a `BatchLogRecordProcessor` whose `Timer.periodic` stays alive after `main()` returns. That timer is what was parking the Dart isolate in `Dart_RunLoop` and preventing process exit.

## The symptom (and why the issue title is slightly off)

The issue says `await OTel.shutdown()` "never returns". Running the user's exact repro locally clarified that `await` *does* return and any post-`await print` runs — what hangs is **process exit**. The leaked `Timer.periodic` on the LoggerProvider's batch processor matches the issue's stack trace (main thread parked in `Dart_RunLoop`, dart:io `EventHandler` thread parked on `kevent`, `DartWorker` threads in `_pthread_cond_wait`).

Before:
```
$ time dart run repro.dart
[2026-05-10 13:23:02.049] starting
[2026-05-10 13:23:02.071] initialized
[2026-05-10 13:23:02.079] span ended
[2026-05-10 13:23:02.135] forceFlush returned
[2026-05-10 13:23:02.140] past shutdown      ← print does run
^C                                            ← only kill -9 makes it exit
```

After:
```
$ time dart run repro.dart
... (same prints) ...
dart run repro.dart  2.10s user 0.20s system 126% cpu 1.815 total   ← exits
```

## Why no existing test caught this

The existing in-process `OTel.shutdown` tests can't catch isolate-stays-alive bugs: `dart test` tears down the test isolate after each test, so a leaked `Timer.periodic` is masked by the test runner. The new test in this PR spawns a separate Dart CLI binary running the exact repro and asserts it exits naturally within 30s — that's the only way to catch process-exit failures.

## Scope / follow-up

The API doesn't yet expose a `loggerProviders()` enumerator (the way it does `tracerProviders()` and `meterProviders()`), so this fix only shuts down the **default** `LoggerProvider`. Named LoggerProviders created via `OTel.addLoggerProvider(name)` still need to be shut down by the caller. Adding the enumerator to the API is a follow-up so `OTel.shutdown()` can clean those up automatically.

## Commits

Two commits per the project's TDD convention:
1. `test:` — failing test that reproduces the bug end-to-end (spawns CLI, expects natural exit)
2. `fix:` — the one-block addition to `OTel.shutdown()` plus a CHANGELOG entry under `## [1.1.0-beta.1-wip]`

## Test plan

- [x] `dart analyze` clean.
- [x] New test reproduces the bug on `HEAD~1`, passes on `HEAD` (verified locally; ~2s exit time vs. the previous 30s deadline timeout).
- [x] Existing `OTel.shutdown` tests still pass (`dart test test/unit/otel_sdk_test.dart test/unit/sdk_edge_cases_test.dart` — 70 passed).
- [x] Full `tool/test.sh` suite passes in CI.
- [x] Manual verification: run any short-lived CLI consumer of the SDK (the dart-otel-reference-demo's swarm runner is the canonical case).


michaelbushe:
Pushed an additional commit (`db6332e`) to address the wider concern of "are all tests actually running":

## What was actually missing

`tool/test.sh` and `tool/coverage.sh` were both invoking:

```
dart test ./test/unit ./test/integration ./test/performance
```

That hand-listed subset silently skipped any test file outside those three directories — including the existing `test/basic_test.dart` at the top level. Add a new test file at the top level or in any sibling directory and CI would happily report "all tests passed" without ever loading it.

The new `test/unit/shutdown_does_not_hang_test.dart` from this PR happens to live under `./test/unit` so it does run (verified in this PR's CI run #25627594831 — both scenarios pass: `✅ ... CLI binary exits within 30s ...` and `✅ ... CLI binary exits cleanly with logs/metrics enabled ...`, totaling 1777 tests). But the underlying bug remains: future tests can land in untested locations.

Both scripts now run `dart test ./test` instead. Browser-only tests under `test/web/` are tagged `@TestOn('browser')` and are auto-skipped on the VM target, so the change is safe; the dedicated browser run still goes through `tool/web_tests.sh` (dart2js + dart2wasm).

## Merge gating ("if any fail, it can't be merged")

That's the **branch protection** setting on `main`, not something a PR can change. Confirmed PR #37 currently shows `mergeStateStatus: BLOCKED` and `reviewDecision: REVIEW_REQUIRED`, so review is enforced. Whether status checks are *also* required — and which specific check names — is in **Settings → Branches → Branch protection rule for `main` → Require status checks to pass before merging**. The check names that should be required:

- `analyze`
- `test (stable)`
- `test (beta)`
- `test-env-vars`
- `test-web`
- `pana`

If those aren't ticked, a failing `test (stable)` won't actually block a merge — only a missing review will. (I couldn't read the protection settings via the API to confirm — that capability is gated.)

PR #39, opened 2026-05-10
chore(release): clarify that pub.dev still prompts before uploading

michaelbushe:
## Summary

Mirrors the same change in the API repo (`dartastic_opentelemetry_api#chore/release-publish-clearer-prompt`). The pre-release summary printed `Publish: yes (auto)`, which read as "non-interactive / no further prompts". `dart pub publish` actually prompts for a final y/N before uploading — the "auto" only meant "the script invokes pub publish for you". Confusing.

After:

```
  Publish:   yes — pub.dev will prompt before uploading
```

Plus the script's own gate now reads `Continue with the local release steps? [y/N]` and a reminder is printed before invoking pub publish.

### DIFF
--- a/tool/release.dart
+++ b/tool/release.dart
@@ -100,10 +100,17 @@ Future<void> main(List<String> args) async {
     );
     _runOrThrow('dart', ['pub', 'get'], silent: true);
     _runOrThrow('dart', ['analyze']);
-    if (!flags.skipTests) {
-      _runOrThrow('dart', ['test']);
+    if (flags.skipTests) {
+      stdout.writeln('(skipping tests — --skip-tests)');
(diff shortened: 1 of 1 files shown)


michaelbushe:
Pushed an additional commit (`85821a5`): when `tool/test.sh` exists and we're on macOS/Linux, the release script invokes that wrapper instead of plain `dart test`. The wrapper downloads/starts the OTel collector for the integration tests, which plain `dart test` hangs on. Falls back to `dart test` when the wrapper isn't present (e.g. the API repo) or on Windows.

Net effect for SDK release flow: just run `dart tool/release.dart` — no need to remember `--skip-tests` and a separate `tool/test.sh` invocation.

PR #46, opened 2026-05-11
fix(web): use user_agent.original instead of removed browser.user_agent

michaelbushe:
## Summary

\`browser.user_agent\` was removed from the OTel browser semconv namespace in favor of the top-level \`user_agent.*\` registry. \`WebResourceDetector\` now emits the navigator's User-Agent string under \`user_agent.original\` (via \`UserAgentResource.userAgentOriginal\` from the API), matching the current spec.

Spec refs:
- https://opentelemetry.io/docs/specs/semconv/registry/attributes/user-agent/
- https://opentelemetry.io/docs/specs/semconv/registry/attributes/browser/ (\`browser.user_agent\` is no longer listed)

## Dependency

Depends on https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/24 which adds the \`UserAgentResource\` enum. Until that's published as \`1.0.0-beta.6\`, this PR uses a \`dependency_overrides\` pointer to the API branch — a TODO in \`pubspec.yaml\` calls out the bump-and-delete that happens once beta.6 lands.

## Notes

- **Breaking on web only**: any backend/dashboard filtering on \`browser.user_agent\` needs to switch to \`user_agent.original\`. Logged under the wip section in CHANGELOG.
- \`browser.vendor\` and \`browser.languages\` were also dropped from current semconv but are kept for now — separate decision.

## Test plan

- [x] \`dart analyze\` — clean.
- [x] \`dart test ./test/unit\` — 1772 unit tests pass.


michaelbushe:
Included in #PR47

PR #53, opened 2026-05-29
Fix SDK initialization after API no-op factory is installed

### LINKED ISSUE #50: OTel.tracerProvider() throws type cast error in tests (APITracerProvider vs TracerProvider)
## Symptom

```
```

## Fix direction

Two options:

robert-northmind:
## Summary

Fixes #50.

The API package can auto-install its no-op `OTelAPIFactory` before the SDK is initialized, for example when API-only/resource-related code runs first. Previously, `OTel.initialize()` treated any installed factory as "already initialized", so the API no-op factory blocked SDK initialization. Later SDK provider access could then fail with API-provider-to-SDK-provider cast errors.

This change lets `OTel.initialize()` replace an installed API no-op factory with the configured SDK factory, while preserving the SDK initialize-first lifecycle for concrete SDK accessors.

## Changes

- Track explicit SDK initialization separately from `OTelFactory.otelFactory` via `OTel.isInitialized`.
- Allow `OTel.initialize()` to replace an installed API no-op `OTelAPIFactory`.
- Preserve clear initialize-first errors for SDK accessors before `OTel.initialize()`, even if the API no-op factory exists.
- Add validation that custom SDK factory creation returns an `OTelSDKFactory`.
- Keep a defensive error for foreign non-API/non-SDK factories.
- Add regression coverage for tracer, meter, logger, and `addTracerProvider` API-first scenarios.

## Notes

This intentionally distinguishes API and SDK lifecycle behavior:

- `OTelAPI.*` may be used before SDK initialization and can install no-op API providers.
- Concrete SDK access through `OTel.*` still requires `OTel.initialize()` first.
- `OTel.initialize()` is responsible for upgrading/replacing the API no-op factory with the configured SDK factory.

This avoids returning temporary/unconfigured SDK providers before initialization.

This follows the same initialize-first principle used by OpenTelemetry Java's global SDK registration: concrete SDK/global access is expected after SDK registration, rather than returning temporary SDK objects that later need to be reconfigured. JS/Python handle early API access with proxy objects that later delegate to the real SDK; this package does not have that proxy layer, so this PR avoids returning unconfigured provisional SDK providers before `OTel.initialize()`.

## Validation

- `dart analyze`
- `dart test test/unit/otel_initialization_error_test.dart test/unit/otel_sdk_test.dart`
- `dart analyze && dart test`

### DIFF
--- a/lib/src/otel.dart
+++ b/lib/src/otel.dart
@@ -38,6 +38,16 @@ import '../dartastic_opentelemetry.dart';
 /// List\<String>, List\<bool>, List\<int> or List\<double>).
 class OTel {
   static OTelSDKFactory? _otelFactory;
+
+  /// Whether [initialize] has been explicitly called by the user.
+  ///
+  /// This is intentionally separate from "an [OTelFactory] is installed":
+  /// the API may auto-install a no-op factory before [initialize] runs (see
+  /// [_ensureSDKFactory]). Only an explicit [initialize] call flips this flag,
+  /// so re-initialization is detected correctly while still allowing the SDK
+  /// to replace the API's no-op factory during first initialization.
+  static bool _userInitialized = false;
+
   static Sampler? _defaultSampler;
   static TimeProvider? _defaultTimeProvider;
 
@@ -80,6 +90,13 @@ class OTel {
 
   /// Default tracer version.
   static String defaultTracerVersion = '1.0.0';
+
+  /// Whether [initialize] has been called (and not undone by [reset]).
+  ///
+  /// This reflects the explicit [initialize] call only, so it can still be
+  /// `false` while the API's no-op factory is installed before the SDK has
+  /// been initialized.
+  static bool get isInitialized => _userInitialized;
 
   /// Initializes the OpenTelemetry SDK with the specified configuration.
   ///
@@ -219,10 +236,28 @@ class OTel {
         resourceAttributes = OTel.attributesFromMap(envResourceAttrs);
       }
     }
-    if (OTelFactory.otelFactory != null) {
+    // Re-initialization is keyed on an explicit prior initialize() call, not
+    // on the mere presence of a factory: the API may have auto-installed a
+    // no-op factory before initialize() runs. That no-op factory is
+    // upgraded/overwritten below.
+    if (_userInitialized) {
       throw StateError(
-        'OTelAPI can only be initialized once. If you need multiple endpoints or service names or versions create a named TracerProvider',
+        'OTel.initialize() can only be called once. For additional telemetry '
+        'pipelines, create named tracer, meter, or logger providers instead.',
       );
+    }
+    final installedFactory = OTelFactory.otelFactory;
+    if (installedFactory is OTelSDKFactory) {
+      throw StateError(
+        'An SDK OpenTelemetry factory (${installedFactory.runtimeType}) is '
+        'already installed. OTel.initialize() can only install the global SDK '
+        'factory once. Call OTel.reset() first if this is a test.',
+      );
+    }
+    if (installedFactory != null && installedFactory is! OTelAPIFactory) {
+      // A foreign, caller-supplied factory is installed; refuse rather than
+      // silently discard it.
+      throw _foreignFactoryError(installedFactory);
     }
 
     if (endpoint.isEmpty) {
@@ -247,11 +282,25 @@ class OTel {
     // Initialize logging from environment variables if needed
     initializeLogging();
 
-    OTelFactory.otelFactory = factoryFactory(
+    // Install the fully-configured SDK factory, overwriting any no-op API
+    // factory installed before now. Using a fresh instance discards any no-op
+    // providers the API factory cached.
+    final createdFactory = factoryFactory(
       apiEndpoint: endpoint,
       apiServiceName: serviceName,
       apiServiceVersion: serviceVersion,
     );
+    // OTelFactoryCreationFunction returns the base OTelFactory type, but SDK
+    // initialization requires the concrete SDK factory implementation.
+    if (createdFactory is! OTelSDKFactory) {
+      throw StateError(
+        'oTelFactoryCreationFunction must create an OTelSDKFactory, got '
+        '${createdFactory.runtimeType}.',
+      );
+    }
+    OTelFactory.otelFactory = createdFactory;
+    _otelFactory = createdFactory;
+    _userInitialized = true;
 
     if (OTelLog.isDebug()) {
       OTelLog.debug(
@@ -512,7 +561,7 @@ class OTel {
   /// @param schemaUrl Optional URL of the schema defining the attributes
   /// @return A new Resource instance
   static Resource resource(Attributes? attributes, [String? schemaUrl]) {
-    _getAndCacheOtelFactory();
+    _ensureSDKFactory();
     return (_otelFactory as OTelSDKFactory).resource(
       attributes ?? OTel.attributes(),
       schemaUrl,
@@ -533,7 +582,7 @@ class OTel {
   /// @return A new ContextKey instance
   static ContextKey<T> contextKey<T>(String name,
       {bool isTransferable = false}) {
-    _getAndCacheOtelFactory();
+    _ensureSDKFactory();
     return _otelFactory!.contextKey<T>(
       name,
       ContextKey.generateContextKeyId(),
@@ -550,7 +599,7 @@ class OTel {
   /// @param spanContext Optional span context to include in the context
   /// @return A new Context instance
   static Context context({Baggage? baggage, SpanContext? spanContext}) {
-    _getAndCacheOtelFactory();
+    _ensureSDKFactory();
     var context = OTelFactory.otelFactory!.context(baggage: baggage);
     if (spanContext != null) {
       context = context.copyWithSpanContext(spanContext);
@@ -571,6 +620,7 @@ class OTel {
   /// @param name Optional name of a specific TracerProvider
   /// @return The TracerProvider instance
   static TracerProvider tracerProvider({String? name}) {
+    _ensureSDKFactory();
     final tracerProvider = OTelAPI.tracerProvider(name) as TracerProvider;
     // Ensure the resource is properly set
     if (tracerProvider.resource == null && defaultResource != null) {
@@ -604,6 +654,7 @@ class OTel {
   /// @param name Optional name of a specific MeterProvider
   /// @return The MeterProvider instance
   static MeterProvider meterProvider({String? name}) {
+    _ensureSDKFactory();
     final meterProvider = OTelAPI.meterProvider(name) as MeterProvider;
     meterProvider.resource ??= defaultResource;
     return meterProvider;
@@ -630,6 +681,7 @@ class OTel {
     Resource? resource,
     Sampler? sampler,
   }) {
+    _ensureSDKFactory();
     final sdkTracerProvider = OTelAPI.addTracerProvider(name) as TracerProvider;
     sdkTracerProvider.resource = resource ?? defaultResource;
     sdkTracerProvider.sampler = sampler ?? _defaultSampler;
@@ -695,7 +747,7 @@ class OTel {
     String? serviceVersion,
     Resource? resource,
   }) {
-    _getAndCacheOtelFactory();
+    _ensureSDKFactory();
     final mp = _otelFactory!.addMeterProvider(
       name,
       endpoint: endpoint,
@@ -736,6 +788,7 @@ class OTel {
   /// @param name Optional name of a specific LoggerProvider
   /// @return The LoggerProvider instance
   static LoggerProvider loggerProvider({String? name}) {
+    _ensureSDKFactory();
     final logProvider = OTelAPI.loggerProvider(name) as LoggerProvider;
     logProvider.resource ??= defaultResource;
     return logProvider;
@@ -760,7 +813,7 @@ class OTel {
     String? serviceVersion,
     Resource? resource,
   }) {
-    _getAndCacheOtelFactory();
+    _ensureSDKFactory();
     final lp = _otelFactory!.addLogProvider(name,
         endpoint: endpoint,
         serviceName: serviceName,
@@ -879,7 +932,7 @@ class OTel {
   /// @param parent The parent SpanContext
   /// @return A new child SpanContext
   static SpanContext spanContextFromParent(SpanContext parent) {
-    _getAndCacheOtelFactory();
+    _ensureSDKFactory();
     return OTelFactory.otelFactory!.spanContextFromParent(parent);
   }
 
@@ -889,7 +942,7 @@ class OTel {
   ///
   /// @return An invalid SpanContext instance
   static SpanContext spanContextInvalid() {
-    _getAndCacheOtelFactory();
+    _ensureSDKFactory();
     return OTelFactory.otelFactory!.spanContextInvalid();
   }
 
@@ -902,7 +955,7 @@ class OTel {
   /// @param attributes Attributes to associate with the event
   /// @return A new SpanEvent instance with the current timestamp
   static SpanEvent spanEventNow(String name, Attributes attributes) {
-    _getAndCacheOtelFactory();
+    _ensureSDKFactory();
     return spanEvent(name, attributes, DateTime.now());
   }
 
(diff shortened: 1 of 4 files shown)


robert-northmind:
Hi, the failing `test (stable)` job appears to be an unrelated flaky timeout in `test/unit/trace/export/otlp/otlp_grpc_span_exporter_edge_test.dart`:

`OtlpGrpcSpanExporter edge cases broken protocol server exercises generic error handling`
Our new initialization tests passed. Could you please rerun the failed job?

benjaben:
@robert-northmind 

One small thing:

`_factoryInstalledBeforeInitializeMessage` and `_sdkAccessorBeforeInitializeMessage` share the same branching logic (`OTelAPIFactory` vs other). They differ only in a few words of phrasing. Not a blocker, but could
  be collapsed into one helper that takes a context string:

```
  static String _badFactoryMessage(OTelFactory installed, {required String context}) {
    if (installed is OTelAPIFactory) {
      return '$context The OpenTelemetry API auto-installed its no-op factory...';
    }
    return '$context A non-SDK factory (${installed.runtimeType}) is already installed.';
  }
```

robert-northmind:
@benjaben thanks for the feedback 🤩 update it

michaelbushe:
Am I crazy?  I thought a week ago I commented on this - was it another issue?  I don't see it.  
I tried to say that this (or some other PR?) was papering over an implementation problem: If the API is installed first - which it almost always will be if resource detection is on (the default) - then the  SDK should upgrade from API to SDK (consistent with other platform implementaitons).  So these errors really need to detect that the factory is the API factory and set OTelFactory.otelFactory = otelSDKFactoryFunction().   

Found it: https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/issues/50

I can make these changes myself but we've been given good guidance that I have done enough in this project and it's best of other contributors do the ongoing work instead of me.

robert-northmind:
> Am I crazy? I thought a week ago I commented on this - was it another issue? I don't see it. I tried to say that this (or some other PR?) was papering over an implementation problem: If the API is installed first - which it almost always will be if resource detection is on (the default) - then the SDK should upgrade from API to SDK (consistent with other platform implementaitons). So these errors really need to detect that the factory is the API factory and set OTelFactory.otelFactory = otelSDKFactoryFunction().
> 
> Found it: #50
> 
> I can make these changes myself but we've been given good guidance that I have done enough in this project and it's best of other contributors do the ongoing work instead of me.

Hi @michaelbushe  thanks for your feedback. And it makes sense. 
I'm happy to update and make those changes. But I am out on vacation for another week. After that I can tackle this.

michaelbushe:
Take your time. It's hot. 😅

robert-northmind:
Hey all, I updated this PR and aligned it better with the comment/feedback from @michaelbushe 
Let me know what you think and if this is in a better direction. Thanks

lib/src/otel.dart
```diff
@@ -219,11 +236,29 @@ class OTel {
...
+    // no-op factory before initialize() runs. That no-op factory is
+    // upgraded/overwritten below.
+    if (_userInitialized) {
+      throw StateError(
+        'OTel.initialize() can only be called once. For additional telemetry '
+        'pipelines, create named tracer, meter, or logger providers instead.',
+      );
+    }
+    final installedFactory = OTelFactory.otelFactory;
+    if (installedFactory is OTelSDKFactory) {
```
michaelbushe, line 250: To made this harder to discern, all the factories extend `OTelAPIFactory` so we can't `is` against the declared types `OTelAPIFactory` nor `OTelSDKFactory` since they match all the factories, which makes some of your code unreachable.

The factoryFunction does a lot of work here.  If an SDK extension has its own factory (not required) it passes it into the SDK via `super.initialize`. Hence, the creation of all SDK factories collapse to this one point. 

Since we must upgrade the API factory we only need to protect against double-SDK init here (for any/all SDKs):
```dart
if (OTelFactory.otelFactory != null && OTelFactory.otelFactory.runtimeType !=  OTelAPIFactory) {
    throw StateError(...);  /no double-inits
}
```

It's cleaner to be explicit about it.  We can add this to the API's `OTelAPIFactory`:
`bool get isAPIFactory => true; `

Then the `OTelSDKFactory` overrides it (and extenders inherit):
`bool get isAPIFactory => false; `

Then this becomes:
```dart
/// Avoid double SDK init 
if (OTelFactory.otelFactory != null && !OTelFactory.otelFactory.isAPIFactory) {
    throw StateError(...); 
}
// Flow down to create the proper SDK factory.  
```

I believe this means that no other changes are necessary - the rest of the code can stay the same.

lib/src/otel.dart
```diff
@@ -81,6 +91,13 @@ class OTel {
   /// Default tracer version.
   static String defaultTracerVersion = '1.0.0';
 
+  /// Whether [initialize] has been called (and not undone by [reset]).
+  ///
+  /// This reflects the explicit [initialize] call only, so it can still be
+  /// `false` while the API's no-op factory is installed before the SDK has
+  /// been initialized.
+  static bool get isInitialized => _userInitialized;
+
```
michaelbushe, line 100: `OTelSDKFactory` is not the only other possible factory. The API is open for further extension. Flutterrific has it's own `FlutterSDKFactory`. SDK consumers can make their own factory and do their own custom OTel tricks.  I hope we gain improvements by seeing what other devs do with their factories.

Claude likes the userInitialized logic but I don't think it's right - I had a Claude-authored PR that was similar that I mentioned in issue 50. It smells bad.  All the factories (see my comment below) have to ask - did you initialize? Did you? 
image


This problem is mitigated somewhat by the `factoryFactory` param that SDK extensions with their own factory must pass into the SDK's `initialize()`, but still this is more complicated than necessary. See below.

lib/src/otel.dart
```diff
@@ -219,11 +236,29 @@ class OTel {
         resourceAttributes = OTel.attributesFromMap(envResourceAttrs);
       }
     }
-    if (OTelFactory.otelFactory != null) {
+    // Re-initialization is keyed on an explicit prior initialize() call, not
+    // on the mere presence of a factory: the API may have auto-installed a
+    // no-op factory before initialize() runs. That no-op factory is
+    // upgraded/overwritten below.
+    if (_userInitialized) {
```
michaelbushe, line 243: LIke above, _userInitialized introduces complications and is not correct since the you have to ask if any of the possible factories have initialized, even the ones that are unknown.  That's what the null-check does.

lib/src/otel.dart
```diff
@@ -1274,24 +1327,44 @@ class OTel {
...
-      return _otelFactory!;
+  /// Ensures an [OTelSDKFactory] is installed as the global factory.
+  ///
+  /// The API may auto-install a no-op [OTelAPIFactory] before [initialize]
+  /// runs. [initialize] is responsible for replacing that no-op factory with
+  /// the configured SDK factory. SDK accessors do not create a temporary SDK
+  /// factory before initialization, because any provider returned from such a
+  /// factory would not automatically become configured when [initialize]
+  /// replaces the global factory later.
+  static OTelSDKFactory _ensureSDKFactory() {
```
michaelbushe, line 1342: This is a better method name but we should update the API and Flutterrific to use the same name for consistency.

lib/src/otel.dart
```diff
@@ -1274,24 +1327,44 @@ class OTel {
...
+    final installedFactory = OTelFactory.otelFactory;
```
michaelbushe, line 1343: Given my above comment, I don't think method has to change at all.  The original does the right thing:
```dart
  static OTelFactory _getAndCacheOtelFactory() {
    if (_otelFactory != null) {
      return _otelFactory!;
    }
    if (OTelFactory.otelFactory == null) {
      throw StateError('initialize() must be called first.');
    }
    return _otelFactory = OTelFactory.otelFactory! as OTelSDKFactory;
  }
```
It either:
- returns the cached factory
- throws if not initialized, or
- the last line can only be reached during SDK initialize() before `OTelFactory.otelFactory = factoryFactory()`, in which case the SDK acts like the API - it installs itself in the interim and then gets upgraded to the passed in factory (which might be its own) later in `initialize()`

michaelbushe (changes requested):
See if you agree with my comments where I claim this can be simpler.  Thank you for the effort!

lib/src/otel.dart
robert-northmind: What do you prefer here for this PR? Should we just keep the original `_getAndCacheOtelFactory` for now, and add a small follow-up change later to rename it consistently across API/SDK/Flutterrific?

lib/src/otel.dart
robert-northmind: 2 questions here:

1: on the:
> the last line can only be reached during SDK initialize()

I think this is also reached when the API auto-installed its no-op factory and a user calls an SDK accessor before `initialize()`. Like in a case like this:

```dart
api.OTelAPI.tracerProvider(); // API-first, e.g. resource detection
OTel.baggage(); // goes through _getAndCacheOtelFactory => _TypeError: type 'OTelAPIFactory' is not a subtype of type 'OTelSDKFactory' in type cast
```

2: on the:
> the SDK acts like the API - it installs itself in the interim and then gets upgraded

I didn't fully understand this part about the "installs itself in the interim and then gets upgraded".
Did you mean that we should add some interim self-install to the SDK (like your #50 suggestion about wrapping the as failure)? 
So if the goal is "early access works and gets upgraded later," that's what Go/JS/Python does with their delegating proxies. But their interim objects can manage being created first and updated later by a later init-call via delegation. 
Fo us, in the current setup, installing a default-configured SDK factory as the interim would behave differently.
Any user created provider in the interim window would use the default endpoint/service name permanently. 

So if we want a "works before init" behavior, then I think we should do some proxy/delegate approach.

And if we want a: "require init before any SDK accessors", then I think we should check that here and return some error like: "OTel.initialize() must be called first"

I think something like this should work:
```dart
static OTelSDKFactory _getAndCacheOtelFactory() {
    if (_otelFactory != null) {
        return _otelFactory!;
    }
    final installedFactory = OTelFactory.otelFactory;
    if (installedFactory == null || installedFactory is! OTelSDKFactory) {
        throw StateError('OTel.initialize() must be called first.');
    }
    return _otelFactory = installedFactory;
}
```

And one more thing. Updating this helper alone doesn't cover the actual stack trace from #50. 
So the tracerProvider(), meterProvider(), loggerProvider(), and addTracerProvider() all cast OTelAPI results directly (... as TracerProvider) without going through the `_getAndCacheOtelFactory`, so they would still throw the raw "APITracerProvider is not a subtype of ..." TypeError before initialize(). I'd add a `_getAndCacheOtelFactory();` call at the top of those four accessors so they fail with the same clear StateError instead.

robert-northmind:
Hi @michaelbushe , thanks for the review and all feedback. Really appreciate it. This is also a good way to learn more about OTel :) Spent some time in the Otel docs now :) 

On the `_userInitialized / isInitialized`, good point. I will remove them. (I didn't know this was a claude-pattern :D interesting. will keep an eye on this. I was actually using a bit of claude and gtp5.5 during this development).

Before I update this PR with more changes, I just wanted to run by my thoughts/plans with you and check in.
So on the `protect against double-SDK init`-logic. Yeah, I like your approach. thanks for the tip.
Could we for now go with the:

```dart
if (OTelFactory.otelFactory != null && OTelFactory.otelFactory.runtimeType !=  OTelAPIFactory) {
    throw StateError(...);  /no double-inits
}
```

Or do you think it makes sense to add the `bool get isAPIFactory => true;` to the API (and default to false for others)?
Or can we for now just leave the check as you suggested and do:
`if (OTelFactory.otelFactory != null && OTelFactory.otelFactory.runtimeType !=  OTelAPIFactory)`?

One other consistency note: 
whichever check we go with, I would use the same one in both `initialize()` and `_getAndCacheOtelFactory`, so the two guards can never disagree about whether the SDK is initialized. I would suggest "is OTelSDKFactory" for both. I know you mentioned it doesn't work against these types, and that's true for is OTelAPIFactory (everything matches it since all factories extend it), but the other direction does discriminate: I verified that a plain OTelAPIFactory instance is not an OTelSDKFactory, while OTelSDKFactory subclasses like OTelFlutterFactory are. It also behaves identically to your isAPIFactory getter for all current factories, so we could swap to the getter later with no behavior change. 

And I left some questions in your other comments.
Thanks again.

michaelbushe:
@robert-northmind Sorry I am so slow lately, I'm headlong on the Dartastic.io launch, which should be this week or next, before FlutterCon. 

I solved this for Dartastic.io in a fork here:
https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/issues/50#issuecomment-4934970955

```dart
    // The API auto-installs its no-op OTelAPIFactory when API-only code runs
    // before the SDK initializes (per spec). That no-op must not block SDK
    // initialization — replace exactly it and nothing else. The runtimeType
    // check is deliberate: every factory (OTelSDKFactory, FlutterSDKFactory,
    // custom consumer factories) extends OTelAPIFactory, so an `is` check
    // would match them all; only the pure auto-installed no-op is replaceable.
    final existingFactory = OTelFactory.otelFactory;
    if (existingFactory != null &&
        existingFactory.runtimeType != OTelAPIFactory) {
      throw StateError(
        'OTelAPI can only be initialized once. If you need multiple endpoints or service names or versions create a named TracerProvider',
      );
    }
    if (existingFactory != null && OTelLog.isDebug()) {
      OTelLog.debug(
        'OTel.initialize: replacing the auto-installed no-op API factory '
        'with the SDK factory. API objects obtained before initialize() '
        'remain no-ops.',
      );
    }
    
    .... /// Then each provider gets....
   
   static TracerProvider tracerProvider({String? name}) {
    // Fail fast with a clear error pre-initialize; without this guard the
    // OTelAPI call below auto-installs the no-op API factory and the cast
    // throws an opaque APITracerProvider-is-not-TracerProvider TypeError (#50).
    _getAndCacheOtelFactory();
....

```

As I just mentioned in the issue, feel free to incorporate it.  I think we are on the same wavelength now.

> 
> Or do you think it makes sense to add the `bool get isAPIFactory => true;` to the API (and default to false for others)?

Yes. It's dead simple and clear. It's an improvement, we can do it in a follow-up and use `existingFactory.runtimeType != OTelAPIFactory` for now - better, make that an isSDKFactory getter which we can !isAPIFactory when we update the API with that call (and override it).

> One other consistency note: whichever check we go with, I would use the same one in both `initialize()` and `_getAndCacheOtelFactory`, so the two guards can never disagree about whether the SDK is initialized. I would suggest "is OTelSDKFactory" for both. 

Yes, consistency is best. Let's make it dirt simple and use `is OTelSDKFactory`, as above, which we can upgrade to !isAPIFactory later.  This would also allow SDKs to work with other APIs implementations. I'd be shocked if someone delivered one but the system is open to extension.  It's nice to keep it open - who knows what other folks will do with it.

My branch looks like so:
```dart
  static OTelFactory _getAndCacheOtelFactory() {
    if (_otelFactory != null) {
      return _otelFactory!;
    }
    final installed = OTelFactory.otelFactory;
    // An installed factory that is not SDK-capable is the API's auto-installed
    // no-op (API-only code ran first) — the SDK still isn't initialized, and
    // saying so beats the `as` TypeError the cast would produce (#50).
    if (installed == null || installed is! OTelSDKFactory) {
      throw StateError('initialize() must be called first.');
    }
    return _otelFactory = installed;
  }

``` 

Give me a week to focus on my launch and then I'll help out the FaroFlutter bug in the other issue.

michaelbushe:
@robert-northmind  Thank you for the deep and extended work on this. Your analysis shaped how I understood the API and the SDK.  I released API beta.8/beta.9, which add OTelFactory.isAPIFactory as discussed. There are other fixes the team requested/added to the API, and we're all delivering products on this and time matter and I made us thrash on this enough already.  I think you'd rather see a solution rather than iterate more.  The important part here is mindshare and you learned a lot ☺️.  So I'm going to land my version rather than iterate here further. I'm adapting several of your tests into it and crediting you as co-author. I value your review on the new PR.  Thanks!

michaelbushe:
Closing superceded by #62

robert-northmind:
Hi @michaelbushe, I'm sorry too for my slow response here 😄 
I am catching up now. Thanks for all the comments and thoughts you left. Makes it easy to follow your thought process. 
And thanks for wrapping this up 👍

PR #54, opened 2026-06-07
feat(trace): configurable exception handling for withSpan

### LINKED ISSUE #51: [FEATURE] Configurable exception handling for Tracer.withSpan / withSpanAsync
**Is your feature request related to a problem? Please describe.**

While integrating `dartastic_opentelemetry` into the Grafana Faro Flutter SDK, we ran into a case where `Tracer.withSpan` and `Tracer.withSpanAsync` currently do the right default thing, but do not allow SDK/library authors to customize exception handling.

Today, when the callback throws, Dartastic appears to always:

```dart
span.recordException(error, stackTrace: stackTrace);
span.setStatus(SpanStatusCode.Error, error.toString());
rethrow;
```

That default is useful and should stay unchanged. The issue is that there is currently no extension point before the exception is recorded or used as the span status description.

For a mobile RUM SDK, raw exception messages can contain PII or sensitive data: request URLs with tokens, user IDs, user-entered text, etc. It might be useful with a way to redact or replace exception details before they become span event attributes or status descriptions.

There are also cases where library authors may want to opt out of automatic exception recording or status updates because their caller handles the error and records telemetry explicitly.

**Describe the solution you'd like**

Add optional exception handling options to withSpan and withSpanAsync, preserving today’s default behavior:

```dart
class SpanExceptionOptions {
  const SpanExceptionOptions({
    this.setStatusOnException = true,
    this.recordException = true,
    this.exceptionSanitizer,
  });

  final bool setStatusOnException;
  final bool recordException;
  final SanitizedSpanException Function(Object error, StackTrace stackTrace)?
      exceptionSanitizer;
}

class SanitizedSpanException {
  const SanitizedSpanException({
    required this.type,
    required this.message,
    this.stackTrace,
    this.statusDescription,
  });

  final String type;
  final String message;
  final StackTrace? stackTrace;
  final String? statusDescription;
}
```

Suggested behavior:

Defaults preserve today’s behavior: record exception, set status to error, rethrow.
- If recordException is false, Dartastic skips automatic exception recording.
- If setStatusOnException is false, Dartastic skips automatic status updates.
- If exceptionSanitizer is provided, Dartastic calls it before recording the exception or setting status.
- If the sanitizer throws, Dartastic still marks the span failed with a generic description such as exception sanitizer failed.
- The original exception is always rethrown.

Example usage:

```dart
await tracer.withSpanAsync(
  span,
  () async {
    return await doWork();
  },
  exceptionOptions: SpanExceptionOptions(
    exceptionSanitizer: (error, stackTrace) {
      return SanitizedSpanException(
        type: error.runtimeType.toString(),
        message: redact(error.toString()),
        stackTrace: sanitizeStackTrace(stackTrace),
        statusDescription: 'sanitized exception',
      );
    },
  ),
);
```

**Describe alternatives you've considered**

Consumers can avoid withSpan / withSpanAsync and manually write try/catch/finally around every span. That works, but it means wrappers such as Faro cannot use Dartastic’s convenience APIs without risking raw exception data being recorded before sanitization.

Another option is to add overloads or parameters directly to recordException, for example explicit sanitized type/message/stack fields. That may still be useful internally, because recordException(error) currently derives exception.type and exception.message from the raw object.

The important part for RUM SDKs is that raw error.toString() should not be required to flow into the span event or status description.

**OpenTelemetry Specification Reference**

The OpenTelemetry exception semantic convention describes the usual catch/record/set-status/rethrow shape:

https://opentelemetry.io/docs/specs/otel/trace/exceptions/
The proposal keeps Dartastic’s default behavior aligned with that pattern, while making the automatic behavior configurable.

**Additional context**

OpenTelemetry Python exposes similar controls on span context-manager helpers:
- `record_exception`
- `set_status_on_exception`

These are available on `start_as_current_span(...)` and `use_span(...)`, both defaulting to `true`, so default behavior remains automatic exception recording/status setting while advanced users can opt out.

Reference:

https://opentelemetry-python.readthedocs.io/en/stable/api/trace.html

Other OTel SDKs vary. JavaScript examples usually catch exceptions manually and call `span.recordException(...)` / `span.setStatus(...)`. Java `@WithSpan` records thrown exceptions through instrumentation, but its public annotation options focus on span name/kind/context, not exception policy.

So the boolean options proposed here are close to Python’s existing API. The sanitizer hook is the additional piece useful for mobile/RUM privacy use cases.

This likely starts in `dartastic_opentelemetry`, where the current automatic record/status behavior lives. If the public APITracer interface needs to accept these options too, a companion change may be needed in `dartastic_opentelemetry_api`.

robert-northmind:
Closes #51

## Summary
- Add `SpanExceptionOptions` with independent `recordException` and `setStatusOnException` controls plus an optional sanitizer hook.
- Wire global defaults through `OTel.initialize`, `TracerProvider`, and per-call overrides on the `withSpan` / active-span helper methods.
- Document the new API in README and CHANGELOG and cover merge/global/per-call behavior in unit tests.

## Validation
- `dart analyze --fatal-infos lib test`
- `dart test test/unit/trace/span_exception_options_test.dart`
- `dart test test/unit/trace/tracer_methods_test.dart test/unit/trace/tracer_method_test.dart test/unit/trace/span_lifecycle_test.dart`

Note: the adjacent tracer test bundle logs expected `localhost:4318` connection refused messages when no local OTLP collector is running; the tests still pass.


benjaben:
@robert-northmind looks pretty solid!

Just a few things maybe we could clarify:

Is it worth adding a test for `startActiveSpanAsync`?
`OTel.withSpan` has a forwarding test but `OTel.withSpanAsync` does not.
`addTracerProvider` with explicit spanExceptionOptions test is missing (maybe not needed?).

robert-northmind:
Hey @benjaben , great feedback. appreciate it ☺️ definitely makes sense to add them. I just updated the PR with them included.

robert-northmind:
I added some more test now. feel free to review again. thanks

robert-northmind:
@michaelbushe if you have time, could you have a look at this PR also. It would be interesting for our FaroFlutter SDK with this functionality. But it is not urgent. thanks

michaelbushe (approved):
Reviewed in depth — approving. This is careful work.

**Verified:**
- The no-leak guarantee holds end-to-end: `recordException` merges caller attributes *over* the derived ones (`copyWithAttributes`), so the sanitized `exception.type`/`exception.message` genuinely replace the raw values, and a null sanitized stackTrace results in no `exception.stacktrace` attribute at all.
- Sanitizer-throws path fails safe: nothing recorded, generic status description, original exception still rethrown.
- Defaults preserve existing behavior exactly; `mergeWith` per-call-over-global semantics are well documented, including the honest caveat that a null sanitizer override can't clear a global one.
- Merged onto current main locally (post-#60, API beta.9): clean merge, `dart analyze` clean, all 18 new tests pass. CI green on the branch.

**Non-blocking nits:** new files say Copyright 2025 (we're in 2026); the CHANGELOG entry is one very long paragraph.

Thanks @robert-northmind — this fixes the FaroFlutter PII-redaction gap (#51) properly.

PR #59, opened 2026-07-10
feat: implement BatchSpanProcessor environment variables

### LINKED ISSUE #58: [FEATURE] Implement Batch Span Processor configuration variables
Ref #20 

Implementation of the high-priority Batch Span Processor env vars.

**Variables to implement:**
* `OTEL_BSP_SCHEDULE_DELAY` - Delay between exports (default: 5000ms)
* `OTEL_BSP_EXPORT_TIMEOUT` - Max export time (default: 30000ms)
* `OTEL_BSP_MAX_QUEUE_SIZE` - Max queue size (default: 2048)
* `OTEL_BSP_MAX_EXPORT_BATCH_SIZE` - Max batch size (default: 512)

**Implementation Notes:**
* Integrate with existing `BatchSpanProcessor`
* Parse integer and duration values
* Validate reasonable ranges
* Document performance implications

harshitt13:
Closes #58

### DIFF
--- a/lib/src/otel.dart
+++ b/lib/src/otel.dart
@@ -457,11 +457,7 @@ class OTel {
 
         spanProcessor = BatchSpanProcessor(
           exporters.length == 1 ? exporter : CompositeExporter(exporters),
-          const BatchSpanProcessorConfig(
-            maxQueueSize: 2048,
-            scheduleDelay: Duration(seconds: 1),
-            maxExportBatchSize: 512,
-          ),
+          BatchSpanProcessorConfig.fromEnvironment(),
         );
       }
       // If exporterType == 'none', spanProcessor remains null and no processor is added
--- a/lib/src/trace/export/batch_span_processor.dart
+++ b/lib/src/trace/export/batch_span_processor.dart
@@ -7,6 +7,7 @@ import 'dart:collection';
 import 'package:dartastic_opentelemetry_api/dartastic_opentelemetry_api.dart';
 import 'package:synchronized/synchronized.dart';
 
+import '../../environment/otel_env.dart';
 import '../span.dart';
 import '../span_processor.dart';
 import 'span_exporter.dart';
@@ -30,21 +31,49 @@ class BatchSpanProcessorConfig {
   final Duration exportTimeout;
 
   /// Creates a new configuration for a [BatchSpanProcessor].
-  ///
-  /// [maxQueueSize] The maximum number of spans that can be queued for export. Default is 2048.
-  ///    If this limit is reached, additional spans will be dropped.
-  /// [scheduleDelay] The time interval between two consecutive exports. Default is 5 seconds.
-  ///    This controls how frequently batches are sent to the exporter.
-  /// [maxExportBatchSize] The maximum number of spans to export in a single batch. Default is 512.
-  ///    This helps control resource usage during export operations.
-  /// [exportTimeout] The maximum time to wait for an export operation to complete. Default is 30 seconds.
-  ///    After this time, export operations will be considered failed.
   const BatchSpanProcessorConfig({
     this.maxQueueSize = 2048,
     this.scheduleDelay = const Duration(milliseconds: 5000),
     this.maxExportBatchSize = 512,
     this.exportTimeout = const Duration(seconds: 30),
   });
+
+  /// Creates a configuration by reading from environment variables via [OTelEnv].
+  /// Falls back to standard OTel defaults if variables are missing or invalid.
+  factory BatchSpanProcessorConfig.fromEnvironment() {
+    final env = OTelEnv.getBspConfig();
+
+    var queueSize = (env['maxQueueSize'] as int?) ?? 2048;
+    var batchSize = (env['maxExportBatchSize'] as int?) ?? 512;
+
+    final scheduleDelay = env['scheduleDelay'] is Duration &&
+            (env['scheduleDelay'] as Duration).inMilliseconds > 0
+        ? env['scheduleDelay'] as Duration
+        : const Duration(milliseconds: 5000);
+    final exportTimeout = env['exportTimeout'] is Duration &&
+            (env['exportTimeout'] as Duration).inMilliseconds > 0
+        ? env['exportTimeout'] as Duration
+        : const Duration(milliseconds: 30000);
+
+    // --- Validation Logic ---
+    if (queueSize <= 0) {
+      queueSize = 2048;
+    }
+    if (batchSize <= 0) {
+      batchSize = 512;
+    }
+    // Spec rule: maxExportBatchSize must be less than or equal to maxQueueSize
+    if (batchSize > queueSize) {
+      batchSize = queueSize;
+    }
+
+    return BatchSpanProcessorConfig(
+      maxQueueSize: queueSize,
+      maxExportBatchSize: batchSize,
+      scheduleDelay: scheduleDelay,
+      exportTimeout: exportTimeout,
+    );
+  }
 }
 
 /// A [SpanProcessor] that batches spans before export.
--- a/lib/src/trace/export/batch_span_processor.dart
+++ b/lib/src/trace/export/batch_span_processor.dart
@@ -17,6 +17,13 @@ import 'span_exporter.dart';
 /// This class configures how the batch span processor behaves, including
 /// queue size limits, export scheduling, and batch size parameters.
 class BatchSpanProcessorConfig {
+  /// Stand-in for "no limit": `OTEL_BSP_EXPORT_TIMEOUT=0` means no timeout
+  /// per spec, represented as the max milliseconds in a 32-bit integer
+  /// (~24.8 days). Web-safe (as microseconds it is far below JS's 2^53
+  /// safe-integer limit, so it behaves identically on VM, dart2js, and wasm)
+  /// and within Dart timer limits for `Future.timeout()`.
+  static const Duration noLimit = Duration(milliseconds: 0x7FFFFFFF);
+
   /// The maximum queue size for spans. After this is reached,
   /// spans will be dropped.
   final int maxQueueSize;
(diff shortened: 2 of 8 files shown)


lib/src/trace/export/batch_span_processor.dart
```diff
@@ -30,21 +31,49 @@ class BatchSpanProcessorConfig {
...
   /// Creates a new configuration for a [BatchSpanProcessor].
-  ///
-  /// [maxQueueSize] The maximum number of spans that can be queued for export. Default is 2048.
-  ///    If this limit is reached, additional spans will be dropped.
-  /// [scheduleDelay] The time interval between two consecutive exports. Default is 5 seconds.
-  ///    This controls how frequently batches are sent to the exporter.
-  /// [maxExportBatchSize] The maximum number of spans to export in a single batch. Default is 512.
-  ///    This helps control resource usage during export operations.
-  /// [exportTimeout] The maximum time to wait for an export operation to complete. Default is 30 seconds.
-  ///    After this time, export operations will be considered failed.
```
michaelbushe, line 41: Please restore this doc. 

Also, please add the spec defaults table to BatchSpanProcessorConfig.fromEnvironment()'s dartdoc (the constants in env_constants.dart already document them individually), and add OTEL_BSP_* to the README's env-var reference — see the OTEL_BLRP_* examples in the logs section for the format.

lib/src/trace/export/batch_span_processor.dart
```diff
@@ -30,21 +31,49 @@ class BatchSpanProcessorConfig {
...
     this.scheduleDelay = const Duration(milliseconds: 5000),
     this.maxExportBatchSize = 512,
     this.exportTimeout = const Duration(seconds: 30),
   });
+
+  /// Creates a configuration by reading from environment variables via [OTelEnv].
+  /// Falls back to standard OTel defaults if variables are missing or invalid.
+  factory BatchSpanProcessorConfig.fromEnvironment() {
+    final env = OTelEnv.getBspConfig();
+
```
michaelbushe, line 72: The > 0 checks deviate from the spec in two ways (sdk-environment-variables.md, types defined in common.md):
  
  1. OTEL_BSP_SCHEDULE_DELAY is a [Duration], and zero is a valid value. The spec says "the value is non-negative. If a negative value is provided, the implementation MUST generate a warning, gracefully ignore the setting and use the default." So OTEL_BSP_SCHEDULE_DELAY=0 (export as fast as possible) should be  honored, not replaced with 5000 ms. The check should be >= 0.
  2. OTEL_BSP_EXPORT_TIMEOUT is a [Timeout], where zero means "no limit". Spec: "Implementations SHOULD interpret timeout zero values as no limit (i.e. infinite). In practice, implementations MAY treat no limit as 'a very long time' and substitute a very large duration." So OTEL_BSP_EXPORT_TIMEOUT=0 should map to a very large  Duration, not 30000 ms.
   /// A stand-in for "no limit": OTEL_BSP_EXPORT_TIMEOUT=0 means no timeout
  /// per spec, which suggests the max milliseconds representable in 32 bits
  /// (~24.8 days) as the substitute.
  static const Duration noLimit = Duration(milliseconds: 0x7FFFFFFF);
  
  I also don't see it used by BatchSpanProcessor. BatchLogRecordProcessor does something similar:
  `exporter.export(...).timeout(_config.exportTimeout)` at batch_log_record_processor.dart:153 and :230
  
  3. Invalid values need a warning, not silence. For negative durations the spec says the SDK "MUST generate a warning, gracefully ignore". For unparseable numerics  it's SHOULD, which the spec tells new implementations to treat as MUST (and this SDK promises to implement all SHOULDs). OTelLog.warn is available and used elsewhere in the SDK. One line per rejection path here  (and in getBspConfig for the int.tryParse failures) covers it.

  Falling back to the defaults is the right. It's the zero-handling and the missing warnings that need adjusting. Would be great to cover these in the tests too (e.g. OTEL_BSP_SCHEDULE_DELAY=0 → 0 ms, =-1 → warn + 5000 ms).

lib/src/trace/export/batch_span_processor.dart
```diff
@@ -30,21 +31,49 @@ class BatchSpanProcessorConfig {
...
+    var batchSize = (env['maxExportBatchSize'] as int?) ?? 512;
+
+    final scheduleDelay = env['scheduleDelay'] is Duration &&
+            (env['scheduleDelay'] as Duration).inMilliseconds > 0
+        ? env['scheduleDelay'] as Duration
+        : const Duration(milliseconds: 5000);
+    final exportTimeout = env['exportTimeout'] is Duration &&
+            (env['exportTimeout'] as Duration).inMilliseconds > 0
+        ? env['exportTimeout'] as Duration
+        : const Duration(milliseconds: 30000);
```
michaelbushe: For the timeout-zero case, a concrete suggestion: the spec's own example of "a very large duration" is the maximum milliseconds representable by a 32-bit integer (~24.8 days):

```dart
/// Stand-in for "no limit": OTEL_BSP_EXPORT_TIMEOUT=0 means no timeout per
/// spec, which suggests the max milliseconds representable in 32 bits.
static const Duration noLimit = Duration(milliseconds: 0x7FFFFFFF);
```

It's web-safe (as microseconds it's far below JS's 2^53 safe-integer limit, so it behaves identically on VM, dart2js, and wasm) and within Dart timer limits for `Future.timeout()`.

Related observation while checking how this field is consumed: `BatchSpanProcessor` stores `exportTimeout` but never applies it — there's no `.timeout(` call anywhere in this file, unlike `BatchLogRecordProcessor`, which does `exporter.export(...).timeout(_config.exportTimeout)` (batch_log_record_processor.dart:153 and :230). Wiring it up the same way could be a follow-up issue rather than scope creep here, but it's worth knowing the env var is currently read and validated without a runtime effect.

test/unit/environment/env_coverage_test.dart
```diff
@@ -72,6 +72,92 @@ class _FixedDetector implements ResourceDetector {
 }
 
 void main() {
+  // =========================================================================
+  // OTelEnv - subprocess tests for getBspConfig
+  // =========================================================================
+  group('OTelEnv.getBspConfig (subprocess)', () {
```
michaelbushe, line 167: These subprocess tests cover `OTelEnv.getBspConfig()` (the string-parsing layer) well, but `BatchSpanProcessorConfig.fromEnvironment()` — where the defaults, the positive-value checks, and the spec's `maxExportBatchSize ≤ maxQueueSize` clamp all live — is never constructed in any test.

Since `Platform.environment` can't be set in-process, the natural fit is a second helper (or a flag on `check_bsp_config.dart`) that constructs `BatchSpanProcessorConfig.fromEnvironment()` in the subprocess and prints the four resolved fields. Cases worth asserting:

- no vars set → 5000 / 30000 / 2048 / 512
- `OTEL_BSP_SCHEDULE_DELAY=0` → 0 ms (valid per spec — see the comment on the `> 0` checks)
- `OTEL_BSP_SCHEDULE_DELAY=-1` → default (and a warning once logging is added)
- `OTEL_BSP_MAX_QUEUE_SIZE=100` + `OTEL_BSP_MAX_EXPORT_BATCH_SIZE=200` → batch clamped to 100

That last one is the Notes-column rule from the spec table ("Must be less than or equal to OTEL_BSP_MAX_QUEUE_SIZE") — implemented in the factory but currently untested.

lib/src/otel.dart
```diff
@@ -425,11 +425,7 @@ class OTel {
 
         spanProcessor = BatchSpanProcessor(
           exporters.length == 1 ? exporter : CompositeExporter(exporters),
-          const BatchSpanProcessorConfig(
-            maxQueueSize: 2048,
-            scheduleDelay: Duration(seconds: 1),
-            maxExportBatchSize: 512,
-          ),
+          BatchSpanProcessorConfig.fromEnvironment(),
```
michaelbushe, line 460: This is a behavior change for every user who sets no env vars: the previous hard-coded config used `scheduleDelay: Duration(seconds: 1)`, and `fromEnvironment()`'s fallback is the spec default of 5000 ms, so default span batching goes from 1s to 5s. Aligning with the spec is likely the right call, but it deserves a CHANGELOG entry under `[1.1.0-beta.7-wip]` noting the change (on mobile, a shorter flush window loses fewer spans when the app is killed, so some users may want to set `OTEL_BSP_SCHEDULE_DELAY=1000` to keep the old behavior).

michaelbushe (changes requested):
Nice first submission.  Some changes requested.  Thank you.

michaelbushe:
Well done, @harshitt13.  Thank you. I'm rebasing (I moved it ahead on you) and merging.

michaelbushe:
Maintainer rebase (with @michaelbushe's go-ahead) — the branch was based on pre-beta.7 main and needed to move past this week's changes (the API's 1.0.0-beta.10 semconv restructure landed via #75). What changed:

- Your three commits are untouched and replayed as-is on current `main` — authorship preserved.
- One housekeeping commit on top: the textual rebase had silently filed your CHANGELOG entries under the *released* `1.1.0-beta.7` heading (and an earlier merge had duplicated them); they now sit once, under `## [1.1.0-beta.8-wip]`.

Verified locally on the rebased branch: `dart analyze` clean, `./tool/test.sh` full suite green (1,866 tests), format clean.

All five review threads from the CHANGES_REQUESTED review check out as addressed in the code — nice work on the `noLimit` web-safety analysis and the `check_bsp_from_env` subprocess coverage. Awaiting re-review.

lib/src/trace/export/batch_span_processor.dart
```diff
@@ -16,6 +17,13 @@ import 'span_exporter.dart';
 /// This class configures how the batch span processor behaves, including
 /// queue size limits, export scheduling, and batch size parameters.
 class BatchSpanProcessorConfig {
+  /// Stand-in for "no limit": `OTEL_BSP_EXPORT_TIMEOUT=0` means no timeout
+  /// per spec, represented as the max milliseconds in a 32-bit integer
+  /// (~24.8 days). Web-safe (as microseconds it is far below JS's 2^53
+  /// safe-integer limit, so it behaves identically on VM, dart2js, and wasm)
+  /// and within Dart timer limits for `Future.timeout()`.
+  static const Duration noLimit = Duration(milliseconds: 0x7FFFFFFF);
```
michaelbushe, line 25: Great comment!

michaelbushe (approved):

PR #62, opened 2026-07-11
fix: API-first usage no longer wedges SDK initialization (#50)

### LINKED ISSUE #50: OTel.tracerProvider() throws type cast error in tests (APITracerProvider vs TracerProvider)

michaelbushe:
Closes #50. Supersedes #53 — see the discussion there.

## The bug

The API auto-installs its no-op `OTelAPIFactory` when API-only code runs before `OTel.initialize()` (per the OTel spec). `initialize()` then refused with "can only be initialized once", and `OTel.tracerProvider()` crashed with the opaque `APITracerProvider is not a subtype of TracerProvider` cast TypeError.

## The fix

- `OTel.initialize()` replaces **exactly** the auto-installed no-op API factory, identified via `OTelFactory.isAPIFactory` (added in API beta.8 for this purpose) — real factories are never silently replaced.
- `OTelSDKFactory` overrides `isAPIFactory` to `false`, as the API ≥ beta.8 contract requires of every real implementation.
- `_getAndCacheOtelFactory()` treats a non-SDK-capable installed factory as not-initialized and throws a clear `OTel.initialize() must be called first.` `StateError` instead of the cast error; `tracerProvider()` / `meterProvider()` / `loggerProvider()` / `addTracerProvider()` guard through it.
- `ObservableResult.observeWithMap` now guards the null factory **before** map conversion — under API beta.9, `toAttributes()` lazily installs the no-op factory, which defeated the uninitialized check (and its regression test).
- Bumps `dartastic_opentelemetry_api` to `^1.0.0-beta.9`: beta.8 adds `isAPIFactory` and no-op replacement in `OTelAPI.initialize()`; beta.9 fixes pre-init lazy no-op installs and makes `Context` re-read the global factory.

## Why now

`^1.0.0-beta.8` on main already resolves to beta.9, and beta.9's lazy-install behavior makes fresh CI runs on current main **red** (e.g. `OTel.attributesFromMap(...)` before `initialize()` now installs the no-op → `initialize()` throws). That is exactly what failed on #54's re-run — those failures are main+beta.9, not #54. This PR turns main green again; #54 just needs a branch update after it lands.

## Tests

Union of the fork's api-first regression tests (`otel_api_first_test.dart`) and the suite contributed by @robert-northmind in #53 (`otel_initialization_error_test.dart`, adapted to factory-state assertions and hermetic initialize flags). Robert is credited as commit co-author.

Full unit suite: **1813 passed, 17 skipped, 0 failures** on API beta.9. `dart analyze` and `dart format` clean.

Thanks @robert-northmind for the investigation and the regression tests — review very welcome.


michaelbushe:
@kevmoo I'm going to merge this to facilitate #54 and get main green again after the API beta.9 update.  Since it's a language trick to force the use of factories, if you have a time please review but it's not necessary.

robert-northmind (commented):
Nice! Looks great! Thanks for completing this

PR #63, opened 2026-07-11
fix: OTel.initialize() upgrades the API's auto-installed no-op factory (#50)

### LINKED ISSUE #50: OTel.tracerProvider() throws type cast error in tests (APITracerProvider vs TracerProvider)

michaelbushe:
## Summary

**SDK main is currently broken against the released `dartastic_opentelemetry_api` 1.0.0-beta.9**: the API now lazily installs its no-op factory whenever API code runs first — including `OTel.initialize()`'s own `OTEL_RESOURCE_ATTRIBUTES` parsing, which runs *before* the once-only guard. Any app setting `OTEL_RESOURCE_ATTRIBUTES` throws `'OTelAPI can only be initialized once'` at startup, and CI's fresh `pub get` hits the same in 9 env subprocess tests on every PR.

TDD: first commit pins the API-first initialization contract red (failing with the production error signature), second turns it green.

## Changes

This is the design from #53's discussion, finalized with the `isAPIFactory` getter added to the API in beta.8 for exactly this purpose (thanks @robert-northmind for driving #53):

- `OTel.initialize()` replaces an installed factory reporting `isAPIFactory == true`; a real factory still triggers the once-only `StateError`.
- `OTelSDKFactory` overrides `isAPIFactory => false`.
- `_getAndCacheOtelFactory` reports `'initialize() must be called first.'` instead of the opaque `as OTelSDKFactory` cast `TypeError` (#50) when only the no-op is installed.
- `observable_result_test`'s null-factory expectation updated to the beta.8+ lazy-install contract.

Closes #50.

## Validation

- `dart analyze` / `dart format --set-exit-if-changed .` — clean
- `./tool/test.sh` — 1817 tests passing, 19 skipped

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.1.0-beta.7-wip]
 
+### Fixed
+- **`OTel.initialize()` upgrades the API's auto-installed no-op factory
+  (#50).** Against `dartastic_opentelemetry_api` >= 1.0.0-beta.8 the API
+  lazily installs its no-op factory whenever API code runs first —
+  including `OTel.initialize()`'s own `OTEL_RESOURCE_ATTRIBUTES` parsing —
+  so initialization threw 'OTelAPI can only be initialized once' and any
+  app setting `OTEL_RESOURCE_ATTRIBUTES` crashed at startup. initialize()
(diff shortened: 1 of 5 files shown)


michaelbushe:
Superseded by #62, which merged the same #50 fix (guard upgrade via isAPIFactory, OTelSDKFactory override, initialize-first errors) with more thorough test coverage. Nothing here adds to it.

PR #66, opened 2026-07-11
feat: implement BatchLogRecordProcessor environment variables and val…

### LINKED ISSUE #65: [FEATURE] Implement Batch LogRecord Processor configuration variables
Ref #20

Implementation of the Batch LogRecord Processor env vars.

**Variables to implement:**
* `OTEL_BLRP_SCHEDULE_DELAY` - Delay between exports (default: 1000ms)
* `OTEL_BLRP_EXPORT_TIMEOUT` - Max export time (default: 30000ms)
* `OTEL_BLRP_MAX_QUEUE_SIZE` - Max queue size (default: 2048)
* `OTEL_BLRP_MAX_EXPORT_BATCH_SIZE` - Max batch size (default: 512)

**Implementation Notes:**
* Integrate with existing `BatchLogRecordProcessor`
* Utilize existing `OTelEnv.getBlrpConfig()` parser
* Validate reasonable ranges and enforce specification rules

harshitt13:
Closes: #65


lib/src/environment/otel_env.dart
```diff
@@ -638,4 +672,33 @@ class OTelEnv {
 
     return null;
   }
+
+  /// Get a positive integer environment variable value.
+  ///
+  /// Returns null when not set, non-numeric, or outside the accepted range.
+  static int? _getPositiveIntEnv(
```
michaelbushe, line 641: The usages of this function always expect a positive value.  If a negative value is sent (yes, should be rare) OTelLog should issue a warn('Illegal negative value for $name').  So doc that it warns on negative values and make it issue an OTelLog warning. 

I can see other violations of "warn on illegal values" in this file which we can improve outside of this PR (or inside if you'd like to do a sweep).

lib/src/environment/otel_env.dart
```diff
@@ -498,42 +490,84 @@ class OTelEnv {
...
+  /// - 'maxQueueSize': int for maximum queue size
+  /// - 'maxExportBatchSize': int for maximum export batch size
+  ///
+  /// Validation rules:
+  /// - Numeric values must be positive integers
+  /// - `maxExportBatchSize` is clamped to `maxQueueSize` per OTel spec
+  /// - Defaults used for validation: queue=2048, batch=512
   static Map<String, dynamic> getBlrpConfig() {
     final config = <String, dynamic>{};
+    const defaultMaxQueueSize = 2048;
```
michaelbushe: These spec defaults now live in three places: here, in `BatchLogRecordProcessorConfig`'s constructor defaults, and in `logs_config.dart`'s `_createProcessor` fallbacks — three homes that can drift apart silently.

The pattern you established in #59 is the right home: give `BatchLogRecordProcessorConfig` a `fromEnvironment()` factory just like your `BatchSpanProcessorConfig.fromEnvironment()`, with the defaults as `static const`s on the config class beside the constructor defaults they must match. Then `OTelEnv.getBlrpConfig()` shrinks back to raw env reading (strings in, parsed values out) and the config class owns its own domain knowledge — defaults, validation, clamping. Bonus: BSP and BLRP end up symmetric, which makes the next signal's env work copy-paste obvious.

Also, please add a CHANGELOG entry under `## [1.1.0-beta.10-wip]` describing the new `OTEL_BLRP_*` validation behavior — it changes what users see for invalid values, so it belongs in the release notes.

lib/src/environment/otel_env.dart
```diff
@@ -498,42 +490,84 @@ class OTelEnv {
...
 
     // Get schedule delay
-    final scheduleDelay = _getEnv(otelBlrpScheduleDelay);
-    if (scheduleDelay != null) {
-      final delayMs = int.tryParse(scheduleDelay);
-      if (delayMs != null) {
-        config['scheduleDelay'] = Duration(milliseconds: delayMs);
-      }
+    final scheduleDelayMs =
+        _getPositiveIntEnv(otelBlrpScheduleDelay, minInclusive: 1);
```
michaelbushe: Warn asymmetry within this method: an invalid `OTEL_BLRP_MAX_QUEUE_SIZE` gets a nice `OTelLog.warn` below, but an invalid `OTEL_BLRP_SCHEDULE_DELAY` or `OTEL_BLRP_EXPORT_TIMEOUT` is dropped silently here — `_getPositiveIntEnv` returns null and nothing tells the operator their value was ignored.  See my comments on `_getPositiveIntEnv` - check my assumption but I think it's safe to say that it should always warn if it's less than `minValueInclusive`.

The spec's sdk-environment-variables page asks implementations to warn on unusable values (and your own queue/batch handling below does exactly that). Simplest fix: move the warn into `_getPositiveIntEnv` itself (it knows the var name and the raw value), so every caller gets consistent diagnostics for free and the `hasMaxQueueSizeEnv`/`hasMaxExportBatchSizeEnv` bookkeeping below can disappear.

lib/src/environment/otel_env.dart
```diff
@@ -498,42 +490,84 @@ class OTelEnv {
...
     }
 
     // Get export timeout
-    final exportTimeout = _getEnv(otelBlrpExportTimeout);
-    if (exportTimeout != null) {
-      final timeoutMs = int.tryParse(exportTimeout);
-      if (timeoutMs != null) {
-        config['exportTimeout'] = Duration(milliseconds: timeoutMs);
+    final exportTimeoutMs =
+        _getPositiveIntEnv(otelBlrpExportTimeout, minInclusive: 1);
```
michaelbushe: `minInclusive: 1` rejects zero for both vars, but zero is meaningful for both per spec — and your own BSP code treats it that way:

- `OTEL_BLRP_SCHEDULE_DELAY=0` is a valid Duration ("export as fast as possible") — `BatchSpanProcessorConfig.fromEnvironment` accepts `>= 0` for the BSP twin.
- `OTEL_BLRP_EXPORT_TIMEOUT=0` is spec type **Timeout**, where zero means "no limit" — BSP maps it to the `noLimit` sentinel rather than rejecting it.

Suggest `minInclusive: 0` for the delay, and explicit `0 → no limit` handling for the timeout, mirroring the BSP semantics so the two batch processors read env vars identically. (This is the kind of thing that becomes free once both go through a shared `fromEnvironment` shape — see the comment above.)

lib/src/environment/otel_env.dart
```diff
@@ -498,42 +490,84 @@ class OTelEnv {
...
-    // Get max export batch size
-    final maxExportBatchSize = _getEnv(otelBlrpMaxExportBatchSize);
-    if (maxExportBatchSize != null) {
-      final size = int.tryParse(maxExportBatchSize);
-      if (size != null) {
-        config['maxExportBatchSize'] = size;
+      // Keep returned queue value internally consistent if batch forced us to clamp
+      // based on an explicitly provided queue value.
+      if (parsedMaxQueueSize != null) {
+        effectiveMaxQueueSize = parsedMaxQueueSize;
```
michaelbushe: This reassignment is a no-op: `effectiveMaxQueueSize` was initialized as `parsedMaxQueueSize ?? defaultMaxQueueSize`, so when `parsedMaxQueueSize != null` it already holds that value — nothing between there and here changes it. The block (and its comment) can be deleted.

Worth saying explicitly though: the `config['maxExportBatchSize'] = effectiveMaxExportBatchSize;` write just above this IS load-bearing and correct — with an env queue of, say, 100 and no batch var set, the consumer's default of 512 would violate the clamp unless the clamped value is written into the config. Nice catch on that case; a short comment on *that* line explaining it would be more valuable than this one.

test/unit/environment/env_coverage_test.dart
```diff
@@ -340,6 +340,79 @@ void main() {
       expect(result.containsKey('maxExportBatchSize'), isFalse);
     });
 
+    test('ignores non-positive scheduleDelay', () async {
```
michaelbushe: Good coverage of the invalid/clamp paths in this suite 👍 — but this particular test asserts the behavior questioned in the otel_env.dart comments: `SCHEDULE_DELAY=0` shouldn't be *ignored*, it's a valid Duration meaning "export as fast as possible" (and `EXPORT_TIMEOUT=0` means "no limit"), matching your BSP `fromEnvironment` semantics. If the zero-handling changes as suggested, this test flips into the two most regression-prone assertions in the file: `0` → `Duration.zero` for the delay, `0` → no-limit for the timeout. Negative values staying ignored-with-warn is exactly right.

michaelbushe (changes requested):
Nice work, please improve according to the comments.  Consistency, simplicity, DRYness and Single Responsibility are always great goals.

lib/src/environment/otel_env.dart
```diff
@@ -498,43 +490,45 @@ class OTelEnv {
...
+  /// Returns a map containing the raw parsed BLRP values from environment
+  /// variables. Domain-level defaults, validation, and clamping belong in
+  /// [BatchLogRecordProcessorConfig.fromEnvironment].
+  ///
+  /// Keys returned (only present when the env var is set and valid):
+  /// - 'scheduleDelay': Duration for the schedule delay
+  /// - 'exportTimeout': Duration for the export timeout
+  /// - 'maxQueueSize': int for maximum queue size
+  /// - 'maxExportBatchSize': int for maximum export batch size
   static Map<String, dynamic> getBlrpConfig() {
```
robert-northmind, line 504: One thought here. I know there is an existing pattern here to return `Map<String, dynamic>` objects. But what if we instead returned like a named dart record. Something like: 
```dart
typedef BlrpEnvironmentValues = ({
  Duration? scheduleDelay,
  Duration? exportTimeout,
  int? maxQueueSize,
  int? maxExportBatchSize,
});
```

Then we get the benefit that we know the types of each entry. and we don't need to use string-keys to find them. I think it would in general make the code a bit safer.
It reduces the need for like `as int?` or `is Duration`? 
Not sure if we wanna go down this path though. 
What are your thoughts on this @michaelbushe ?

lib/src/environment/otel_env.dart
```diff
@@ -498,43 +490,45 @@ class OTelEnv {
...
-    // Get max queue size
-    final maxQueueSize = _getEnv(otelBlrpMaxQueueSize);
-    if (maxQueueSize != null) {
-      final size = int.tryParse(maxQueueSize);
-      if (size != null) {
-        config['maxQueueSize'] = size;
-      }
+    // Get queue and batch sizes — must be positive (>0)
+    final parsedMaxQueueSize =
+        _getPositiveIntEnv(otelBlrpMaxQueueSize, minInclusive: 1);
```
robert-northmind, line 523: One thought here. Does this pub a bit of Blrp domain knowledge in the OTelEnv parsing? Like that a queue needs to be bigger than 0? 
What if the OTelEnv parsing only checks if the env variable exisits and has right type? Like it could warn if an int is a string.

But then the logic to check if the acutal int is bigger than 0, that could happen in the BatchLogRecordProcessorConfig.fromEnvironment()?

lib/src/logs/export/batch_log_record_processor.dart
```diff
@@ -41,6 +48,97 @@ class BatchLogRecordProcessorConfig {
...
+  /// | `OTEL_BLRP_EXPORT_TIMEOUT`        | Timeout  | `30000`  | Export timeout (ms). 0 means no limit.                   |
+  /// | `OTEL_BLRP_MAX_QUEUE_SIZE`        | Integer  | `2048`   | Maximum log record queue size.                           |
+  /// | `OTEL_BLRP_MAX_EXPORT_BATCH_SIZE` | Integer  | `512`    | Maximum batch size. Must be ≤ `MAX_QUEUE_SIZE`.          |
+  ///
+  /// Invalid or out-of-range values emit an [OTelLog.warn] diagnostic and
+  /// fall back to the spec default.
+  factory BatchLogRecordProcessorConfig.fromEnvironment() {
+    final env = OTelEnv.getBlrpConfig();
+
+    var queueSize = (env['maxQueueSize'] as int?) ?? 2048;
```
robert-northmind, line 68: We have now this `2048` defined here and in the constructor in line 46 also, and in a few other places in this file.
should we break that our to a private static default variable?
Like: `static const _defaultQueueSize = 2048;`

lib/src/logs/export/batch_log_record_processor.dart
```diff
@@ -41,6 +48,97 @@ class BatchLogRecordProcessorConfig {
     this.maxExportBatchSize = 512,
     this.exportTimeout = const Duration(seconds: 30),
   });
+
+  /// Creates a configuration by reading `OTEL_BLRP_*` environment variables
```
robert-northmind, line 52: What if we make the `BatchLogRecordProcessorConfig` constructor private, and make all its params required. So that we dont have to provide the defaults there.
And instead, we can only create `BatchLogRecordProcessorConfig` either via the `factory BatchLogRecordProcessorConfig.fromEnvironment()` or a `factory BatchLogRecordProcessorConfig({...});`.

Then in those factories we could do the validation. something like:
```dart
factory BatchLogRecordProcessorConfig({
    int maxQueueSize = defaultQueueSize,
    // ...
  }) {
    if (!_validQueueSize(maxQueueSize)) {
      throw ArgumentError.value(maxQueueSize, 'maxQueueSize');
    }
    return BatchLogRecordProcessorConfig._(
      maxQueueSize: maxQueueSize,
    );
  }

factory BatchLogRecordProcessorConfig.fromEnvironment() {
    final env = OTelEnv.getBlrpConfig();
    final queueSize = _validQueueSize(env.maxQueueSize)
        ? env.maxQueueSize!
        : defaultQueueSize;
    if (env.maxQueueSize != null && !_validQueueSize(env.maxQueueSize)) {
      OTelLog.warn(/* invalid environment value */);
    }
    return BatchLogRecordProcessorConfig(
      maxQueueSize: queueSize,
    );
  }
```

lib/src/environment/otel_env.dart
```diff
@@ -638,4 +632,47 @@ class OTelEnv {
 
     return null;
   }
+
+  /// Get a non-negative integer environment variable value.
+  ///
+  /// Returns null when not set, non-numeric, or outside the accepted range.
+  /// Warns via [OTelLog.warn] when the raw value is present but unusable
+  /// (non-numeric, below [minInclusive], or above [maxInclusive]).
+  static int? _getPositiveIntEnv(
```
robert-northmind, line 641: does this need to be private? if it would be public, we could easily directly test the logic in here. Just a thought

lib/src/logs/export/logs_config.dart
```diff
@@ -189,30 +189,42 @@ class LogsConfiguration {
...
+    return BatchLogRecordProcessor(exporter, processorConfig);
+  }
 
+  /// Builds [BatchLogRecordProcessorConfig] from a BLRP environment config map.
+  ///
+  /// Exposed for testing to validate normalization and spec-rule handling.
+  /// Prefer [BatchLogRecordProcessorConfig.fromEnvironment] for production use.
+  static BatchLogRecordProcessorConfig buildBatchLogRecordProcessorConfig(
+    Map<String, dynamic> blrpConfig,
+  ) {
```
robert-northmind, line 202: I didn't quite follow the need for this one. this is only used for testing right? 
is there some other way to structure this. so that we dont need public api for testing only needs?

robert-northmind (commented):
Hey, looks very good. Thanks for the contribution 
I just left some comments. Let me know what you think. I am also new to this codebase. so i might not be right in all my thoughts ☺️

lib/src/environment/otel_env.dart
harshitt13: Good catch, you're right. I'll update `getBlrpConfig()` to only do type-checking (parse the int, warn if non-numeric) and move the `> 0` validation into `fromEnvironment()`.

lib/src/environment/otel_env.dart
harshitt13: Nice idea, but I'll only change BLRP in this PR to keep scope manageable. We could migrate the others (especially BSP, since it's structurally identical) in a follow-up PR if you think the pattern works well. Thoughts @michaelbushe?

lib/src/logs/export/batch_log_record_processor.dart
harshitt13: I'd make them public rather than private (`defaultMaxQueueSize` instead of `_defaultMaxQueueSize`), since tests and external consumers might want to reference the spec defaults. Does that work for you?

lib/src/logs/export/logs_config.dart
harshitt13: Right, I'll remove it and update the test to validate clamping behavior by constructing `BatchLogRecordProcessorConfig(maxQueueSize: 100, maxExportBatchSize: 200`) directly.

lib/src/logs/export/batch_log_record_processor.dart
harshitt13: The idea of centralizing validation is good, but making the constructor a factory would break `const` construction; the codebase currently uses const `BatchLogRecordProcessorConfig()` in several places (including tests and the `BatchLogRecordProcessor` default). Factory constructors can't be `const` in Dart.

Also, the existing `BatchSpanProcessorConfig` (BSP) uses the same pattern (public `const` constructor + `fromEnvironment()` factory) without private constructors, so changing BLRP alone would create an inconsistency between the two.

Options:

1. Go with your pattern and accept the `const` break (minor, just tests/defaults need updating) for cleaner validation
2. Or keep the public constructor, extract defaults to `static const`, and do validation only in `fromEnvironment()` (consistent with BSP)

What do you prefer?

cc: @michaelbushe

lib/src/environment/otel_env.dart
harshitt13: It's already indirectly tested through `getBlrpConfig()` and `getBspConfig()`; the in-process tests and subprocess tests cover the non-numeric, below-minimum, and above-maximum branches.
It is a genuinely reusable parsing utility, so making it public isn't harmful. I could go either way:

- Make it public (it's a useful utility on a public class)
- Keep it private with `@visibleForTesting` 
- Leave it private since it's already well-covered

What's your preference?

cc: @michaelbushe

lib/src/environment/otel_env.dart
robert-northmind: nice, good idea.

lib/src/logs/export/batch_log_record_processor.dart
robert-northmind: super. sounds great

lib/src/logs/export/batch_log_record_processor.dart
robert-northmind: aha, good point. I didn't consider that. yeah, lets wait from some input from @michaelbushe since he built up most of this logic. But I would be okay going with option 2.

lib/src/environment/otel_env.dart
robert-northmind: My vote would be for the public utility then. I am not a super big fan of the `@visibleForTesting` pattern. But in some places it makes sense.

lib/src/logs/export/logs_config.dart
robert-northmind: nice, thanks

test/unit/environment/helpers/check_blrp_from_env.dart
```diff
@@ -0,0 +1,28 @@
...
+
+  final jsonConfig = <String, dynamic>{
+    'scheduleDelay_ms': config.scheduleDelay.inMilliseconds,
+    'exportTimeout_ms': config.exportTimeout.inMilliseconds,
+    'maxQueueSize': config.maxQueueSize,
+    'maxExportBatchSize': config.maxExportBatchSize,
+    'logs': logs,
+  };
+
+  print(jsonEncode(jsonConfig));
```
yuzurihaaa, line 27: I believe this should be wrapped with `test`. And if the intention is for the default value, then we can do not null assertion.

lib/src/logs/export/batch_log_record_processor.dart
robert-northmind: I think from my point of view, we can go with option 2, and keep the public const constructor. Let's just do that for now.

test/unit/environment/helpers/check_blrp_from_env.dart
robert-northmind: Hey, I think this one is fine as is ☺️

This file isn't really a test, it's a helper script that gets run as a subprocess. There are a bunch of others like it in the same folder (check_blrp_config.dart, check_exporter.dart, etc). 

The real tests are in env_coverage_test.dart, they call runWithEnv(...) on these helpers and then parse the printed json.

I think the reason for that setup is that we can't change env vars inside a running dart process, so we have to spawn a new one with the vars already set. So wrapping it in test would actually break it, the stdout is what the test asserts on.

robert-northmind: But actually, I think there is another way to manage this test.
In upstream main, there is a new API in the `EnvironmentService` which allows tests to override envs. This API: `EnvironmentService.testOverrides`.
So if this PR pulls in latest main, then we could use that and do something like:

```dart
setUp(() {
    // Prevent accidentally reading the developer's real environment.
    EnvironmentService.testOverrides = {};
});

tearDown(() {
    EnvironmentService.testOverrides = null;
});

test('uses BLRP defaults when environment is unset', () {
    EnvironmentService.testOverrides = {};
    final config = BatchLogRecordProcessorConfig.fromEnvironment();
    expect(config.maxQueueSize, 2048);
    expect(config.maxExportBatchSize, 512);
    expect(config.scheduleDelay, const Duration(milliseconds: 1000));
    expect(config.exportTimeout, const Duration(seconds: 30));
});

test('reads BLRP values from environment', () {
    EnvironmentService.testOverrides = {
         'OTEL_BLRP_MAX_QUEUE_SIZE': '100',
        'OTEL_BLRP_MAX_EXPORT_BATCH_SIZE': '50',
    };
    final config = BatchLogRecordProcessorConfig.fromEnvironment();
    expect(config.maxQueueSize, 100);
    expect(config.maxExportBatchSize, 50);
});
```

lib/src/environment/otel_env.dart
michaelbushe: Yes definitely agree.  I'm not a fan of having dynamic anywhere and have tried to stomp it out but there's room for improvement.  @harshitt13 - this has been sitting for a while so I will leave it up to you to change this PR.  

Added issue
https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/issues/97

test/unit/environment/helpers/check_blrp_from_env.dart
harshitt13: Thanks for the suggestion @robert-northmind

lib/src/environment/otel_env.dart
```diff
@@ -5,6 +5,19 @@ import 'package:dartastic_opentelemetry_api/dartastic_opentelemetry_api.dart';
 import 'env_constants.dart';
 import 'environment_service.dart';
 
+/// Raw BLRP environment values parsed by [OTelEnv.getBlrpConfig].
+///
+/// All fields are nullable — `null` means the corresponding env var was
+/// unset or contained a non-numeric value (a warning is logged in that case).
+/// Domain-level validation (e.g. "queue size must be > 0") belongs in
```
michaelbushe, line 12: 💙
For the rest, I created https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/issues/97 for (an)other PR(s).

Please include the env vars in the doc of the fields, like OTEL_BLRP_SCHEDULE_DELAY, OTEL_BLRP_EXPORT_TIMEOUT, etc.

lib/src/logs/export/batch_log_record_processor.dart
michaelbushe: Some extra feedback @robert-northmind :
`throw ArgumentError.value(maxQueueSize, 'maxQueueSize');`
is not OK.  We can only throw from initialize(), otherwise, we do the best we can.  That would be a better case for OTelLog.warn. 
See:
https://opentelemetry.io/docs/specs/otel/error-handling/#basic-error-handling-principles

Keep in mind and I think there's a lot of room for improvement.  For sure many of the 84 spec compliance issues that Fable found fall in this category. (I'll split those into separate issues)

michaelbushe: Agreed on @harshitt13 's approach, which is now implemented.

lib/src/logs/export/batch_log_record_processor.dart
```diff
@@ -36,11 +55,104 @@ class BatchLogRecordProcessorConfig {
...
+    }
+
+    // --- exportTimeout ---
+    // Spec type: Timeout. Zero means "no limit" — substitute a very large
+    // duration. Negative values MUST warn and fall back to default.
+    Duration exportTimeout;
+    if (env.exportTimeout != null) {
+      final timeout = env.exportTimeout!;
+      if (timeout.inMilliseconds == 0) {
+        exportTimeout = noLimit;
```
michaelbushe, line 116: This is such a weird corner case that we should add a OTel.debug to tell confused implementers that we switched out their zero for a large number under the covers.

lib/src/logs/export/batch_log_record_processor.dart
```diff
@@ -36,11 +55,104 @@ class BatchLogRecordProcessorConfig {
...
+              'invalid per spec, using default '
+              '${defaultExportTimeout.inMilliseconds} ms.');
+        }
+        exportTimeout = defaultExportTimeout;
+      }
+    } else {
+      exportTimeout = defaultExportTimeout;
+    }
+
+    // --- Domain validation (moved from OTelEnv) ---
```
michaelbushe: Remove "moved from OTelEnv)". Consider the code a only the current state.

CHANGELOG.md
```diff
@@ -19,6 +19,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   two hand-rolled copies of the same assembly; wire output is unchanged
   (same instrumentation-scope constant, same `OTel.resource(null)`
   fallback, resolved by the caller so the transformer stays a pure leaf).
+- **`BatchLogRecordProcessorConfig.fromEnvironment()`** — new factory
```
michaelbushe: This doesn't need to be in the changelog since (I assume) it's not exported, so it's internal.  Please remove since the CHANGELOG is for only public-facing changes.  It reduces noise on a doc that's read often. Thank you.

michaelbushe (commented):
Real close now, just some tiny nits. Thank you, I'll turn around the next one quickly.

michaelbushe (approved):
Super! Nice work, thanks for your patience @harshitt13

PR #68, opened 2026-07-11
fix: gRPC broken-protocol edge test hang (event-loop starvation) + timeout backstop for discussion

michaelbushe:
Surfaced by running `./tool/test.sh` locally (concurrency 20): `otlp_grpc_span_exporter_edge_test`'s **"broken protocol server exercises generic error handling"** intermittently hangs to the test-timeout ceiling. Not a regression from recent merges — it reproduces on older commits; the coverage/test environment just isn't exercised by CI.

Tagging @robert-northmind, @kevmoo, @harshitt13 — this one has a genuinely interesting Dart-async root cause and an open design question, and I'd value your eyes on it.

## Investigation

I reproduced deterministically with a **16× parallel** harness (16 copies of just this test, forcing the CPU contention that concurrency-20 creates):

| Trigger | Result under 16× contention |
|---|---|
| accept-then-RST raw server (current `main`) | **6/16 hang** to the 30s ceiling |
| connection-refused (this PR) | **16/16 pass**, ~5s each |

Instrumenting the hanging runs showed they stop at `export start` and never reach `export done` — the hang is inside `export()`.

## Root cause: event-loop starvation, not an unenforced timeout

My first hypothesis was "the configured `timeout` isn't enforced," so I added a Dart-level `.timeout()` backstop around the RPC. **It didn't fix it** — the failing runs still hung with the backstop in place. The real mechanism:

The test's raw server **accepts** each TCP connection then immediately **RSTs** it (`socket.destroy()`). That drives the gRPC client into a tight connect → RST → reconnect loop that floods the isolate's **microtask queue**. Microtasks always drain before timers/events, so a sustained microtask flood **starves the event loop** — and Dart `Timer`s can't fire on a starved loop. That's why *neither* gRPC's own `CallOptions` deadline *nor* a Dart `.timeout()` fires: both are Timer-based. A timeout backstop categorically cannot bound event-loop starvation.

Under adequate CPU (single test in isolation) the loop keeps up and the call errors in ~5s, which is why it only flaked under load and why the old workaround was just bumping the ceiling to 90s (PR #36, #41).

## The fix

Replace the adversarial accept-then-destroy server with a **connection-refused** trigger: bind a socket to grab a free port, close it, point the exporter there. Connection-refused fails fast and cleanly — same generic error-handling code path, no reconnect churn, no microtask flood. **16/16 under the same contention.** The test ceiling drops from 90s back to a normal 30s (which now also guards against regression: a hang fails at 30s instead of riding to 90s).

## Open question for reviewers: keep the `.timeout()` backstop?

This PR **also** includes the Dart-level timeout backstop in `OtlpGrpcSpanExporter._tryExport` (visible in the diff). My honest assessment:

- It does **not** fix the hang above (that's starvation).
- Its only remaining value is a healthy-loop case where an RPC hangs but gRPC's own `CallOptions` deadline somehow doesn't fire. I **could not prove** that case exists — gRPC's deadline already produces a proper `GrpcError.deadlineExceeded` for an accept-but-never-respond collector, and in my passing runs I can't distinguish the backstop firing from gRPC's own deadline firing.
- Cost: extra logic on the core export path; it changes the thrown type (`TimeoutException` vs `GrpcError`), which flows differently through the retry loop.

**My lean: drop the backstop, ship the test fix only.** But I'd rather the team decide — if you know a real scenario where gRPC's deadline doesn't fire on a healthy loop, that argues for keeping it. If we drop it, I'll revert the exporter change and the CHANGELOG entry and this becomes a test-only PR.

## Related note (not in this PR)

`tool/test.sh` still runs at `--concurrency=20` while `tool/coverage.sh` was lowered to 10 in #67. Lower contention would make starvation-class flakes rarer but wouldn't fix them. Happy to align test.sh to 10 in a follow-up if wanted.

## Validation

- 16/16 under 16× parallel contention (was 6/16 hanging)
- Full edge-test file: 21/21
- `dart analyze` / `dart format`: clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [1.1.0-beta.7] - 2026-07-11
 
 ### Fixed
+- **`OtlpGrpcSpanExporter.export()` gains a Dart-level timeout backstop.**
+  Previously the configured `timeout` was applied only via gRPC's `CallOptions`
+  deadline. A Dart-level `.timeout()` now also bounds the RPC and tears down the
+  channel on expiry, as defense-in-depth for real-world hangs where a collector
+  accepts a connection then stops responding. **Note (under review):** this does
+  NOT fix the concurrency test hang that prompted it — that was event-loop
+  starvation from the gRPC client's reconnect churn, which no Timer-based bound
+  can fix (see the PR discussion). Reviewers are deciding whether to keep this
+  backstop; if dropped, this entry goes with it.
 - **Debug logging no longer adds a `ConsoleExporter` to the trace pipeline.**
   `OTel.initialize()` used to append a `ConsoleExporter` to the span exporters
   whenever debug logging was enabled (e.g. `OTEL_LOG_LEVEL=debug`/`trace`),
--- a/lib/src/trace/export/otlp/otlp_grpc_span_exporter.dart
+++ b/lib/src/trace/export/otlp/otlp_grpc_span_exporter.dart
@@ -345,7 +345,25 @@ class OtlpGrpcSpanExporter implements SpanExporter {
         );
       }
 
-      final response = await _traceService!.export(request, options: options);
+      // gRPC's CallOptions deadline (set above) can fail to terminate a call
+      // when the connection is torn down mid-flight — e.g. a half-open or
+      // broken collector that accepts then drops the TCP connection — leaving
+      // this await hung well past _config.timeout. This Dart-level backstop
+      // guarantees export() is bounded by the configured timeout regardless of
+      // the gRPC client's connection-state behavior. On timeout we tear the
+      // channel down so the leaked RPC is aborted and the next attempt (if any)
+      // reconnects cleanly.
+      final response =
+          await _traceService!.export(request, options: options).timeout(
+        _config.timeout,
+        onTimeout: () {
+          unawaited(_cleanupChannel());
+          throw TimeoutException(
+            'OtlpGrpcSpanExporter: export exceeded configured timeout',
+            _config.timeout,
+          );
+        },
+      );
       if (OTelLog.isDebug()) {
         OTelLog.debug(
           'OtlpGrpcSpanExporter: Export request completed successfully',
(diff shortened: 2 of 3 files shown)


lib/src/trace/export/otlp/otlp_grpc_span_exporter.dart
```diff
@@ -345,7 +345,25 @@ class OtlpGrpcSpanExporter implements SpanExporter {
...
+      // when the connection is torn down mid-flight — e.g. a half-open or
+      // broken collector that accepts then drops the TCP connection — leaving
+      // this await hung well past _config.timeout. This Dart-level backstop
+      // guarantees export() is bounded by the configured timeout regardless of
+      // the gRPC client's connection-state behavior. On timeout we tear the
+      // channel down so the leaked RPC is aborted and the next attempt (if any)
+      // reconnects cleanly.
+      final response =
+          await _traceService!.export(request, options: options).timeout(
+        _config.timeout,
```
harshitt13, line 358: Let's remove this `.timeout()` backstop. As you suspected, it doesn't solve starvation. Also, it introduces a bug: it throws a `TimeoutException`, which falls into the generic `catch (e)` block below and gets retried. Timeouts aren't supposed to be retried, unlike generic connection drops.

CHANGELOG.md
```diff
@@ -9,6 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ### Fixed
+- **`OtlpGrpcSpanExporter.export()` gains a Dart-level timeout backstop.**
+  Previously the configured `timeout` was applied only via gRPC's `CallOptions`
+  deadline. A Dart-level `.timeout()` now also bounds the RPC and tears down the
+  channel on expiry, as defense-in-depth for real-world hangs where a collector
+  accepts a connection then stops responding. **Note (under review):** this does
+  NOT fix the concurrency test hang that prompted it — that was event-loop
+  starvation from the gRPC client's reconnect churn, which no Timer-based bound
+  can fix (see the PR discussion). Reviewers are deciding whether to keep this
+  backstop; if dropped, this entry goes with it.
```
harshitt13, line 54: And if we're going to drop the backstop, then revert this entry

PR #64, opened 2026-07-11
chore: CNCF headers — Copyright The OpenTelemetry Authors + SPDX (#57)

michaelbushe:
Fixes #57 — replaces the per-file `// Licensed under the Apache License, Version 2.0` / `// Copyright 2025, Michael Bushe, All rights reserved.` header with the standard OpenTelemetry form, per the CNCF copyright-notice policy and the OTel community process doc ("We use 'Copyright The OpenTelemetry Authors' notice form"):

```dart
// Copyright The OpenTelemetry Authors
// SPDX-License-Identifier: Apache-2.0
```

- All hand-written Dart sources converted, including files that had no header and variant headers (stray `Copyright 2026` lines, two mid-file license lines), and the files added by #62.
- Generated `lib/proto/**` left as-is — regenerated from upstream `opentelemetry-proto`; CNCF policy says not to alter third-party notices.
- The three `tool/read_*.dart` scripts keep their `#!/usr/bin/env dart` shebang as the first line, header below.
- `tool/add_header.sh` emits and detects the new header.

Rebased onto main after #62 (which superseded #63).

Mirrors dartastic_opentelemetry_api#46.

## Validation

- `dart analyze` / `dart format --set-exit-if-changed .` — clean
- `./tool/test.sh` — full suite green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #72, opened 2026-07-12
Feature/issue 20 metrics env vars

### LINKED ISSUE #71: [FEATURE] Implement Metrics SDK configuration variables
Ref #20

Implementation of the medium-priority Metrics SDK env vars.

**Variables to implement:**
* `OTEL_METRICS_EXEMPLAR_FILTER` - Exemplar filter (default: "trace_based")
* `OTEL_METRIC_EXPORT_INTERVAL` - Export interval (default: 60000ms)
* `OTEL_METRIC_EXPORT_TIMEOUT` - Export timeout (default: 30000ms)

**Implementation Notes:**
* Integrate with `PeriodicExportingMetricReader`
* Parse duration values for intervals and timeouts
* Implement exemplar filtering logic
* Validate enum values for the filter (`"trace_based"`, `"always_on"`, `"always_off"`)

harshitt13:
Closes #71


lib/src/metrics/export/metrics_sdk_config.dart
```diff
@@ -0,0 +1,32 @@
+// Licensed under the Apache License, Version 2.0
+// Copyright 2025, Michael Bushe, All rights reserved.
```
michaelbushe, line 2: Repo convention for new files is the CNCF header (see #64, which converted the whole tree):

```dart
// Copyright The OpenTelemetry Authors
// SPDX-License-Identifier: Apache-2.0
```

This package is bound for CNCF donation, so my personal copyright lines - recently deleted - should not exist.

lib/src/metrics/export/metrics_sdk_config.dart
```diff
@@ -0,0 +1,32 @@
...
+
+  /// Export all exemplars.
+  alwaysOn,
+
+  /// Export no exemplars.
+  alwaysOff,
+}
+
+/// Resolved metrics SDK configuration derived from environment variables.
+class MetricsSdkConfig {
```
michaelbushe, line 17: A typed config class instead of a `Map<String, dynamic>` bag — this is exactly the direction the env config should be heading (it's what the #66 review asks for on the BLRP side), nice. Two refinements to finish the shape:

1. Move the parsing here as a `MetricsSdkConfig.fromEnvironment()` factory (the `parseMetricsSdkConfig`/`_parseMetricsExemplarFilter`/`_parseDurationMilliseconds` trio currently in `OTelEnv`), following your `BatchSpanProcessorConfig.fromEnvironment()` from #59. `OTelEnv` stays the raw env-reading layer; the config class owns its defaults and validation.
2. `MetricsSdkConfig` and `MetricsExemplarFilter` are returned by a public API but aren't exported from `lib/dartastic_opentelemetry.dart` — as written, callers can receive a type they can't name. Add the barrel export (or make the API internal until it's meant to be public).

lib/src/environment/otel_env.dart
```diff
@@ -467,6 +468,74 @@ class OTelEnv {
...
+    switch (value?.toLowerCase()) {
+      case 'always_on':
+        return MetricsExemplarFilter.alwaysOn;
+      case 'always_off':
+        return MetricsExemplarFilter.alwaysOff;
+      case null:
+      case 'trace_based':
+        return MetricsExemplarFilter.traceBased;
+      default:
+        if (OTelLog.isDebug()) {
```
michaelbushe, line 509: `OTelLog.debug` hides these from anyone not running with debug diagnostics on — but an operator who typo'd `OTEL_METRICS_EXEMPLAR_FILTER=alwayson` is exactly the person who needs to hear about it. The spec asks for a warning on unusable values, and the rest of this file warns (`getBspConfig`, the BLRP work in #66 after review). Please use `OTelLog.isWarn()`/`OTelLog.warn` here and in `_parseDurationMilliseconds` below.

lib/src/metrics/export/metric_config.dart
```diff
@@ -40,6 +40,9 @@ class MetricsConfiguration {
       meterProvider.resource = resource;
     }
 
+    final metricsSdkConfig = OTelEnv.getMetricsSdkConfig();
+    MetricTransformer.setExemplarFilter(metricsSdkConfig.exemplarFilter);
```
michaelbushe, line 44: This is the one structural issue in the PR, and it's worth understanding deeply because it's a spec-layering question, not a style one.

Per the spec, the exemplar **filter** decides whether a measurement is *offered* to the exemplar reservoir **at recording time** — it's an SDK collection policy, not an export-time drop. Applying it in the OTLP transformer means (a) we pay the full cost of collecting exemplars and then throw them away at the wire, and (b) other exporters (console, a future second OTLP endpoint) silently share whatever filter the last `initialize` set, because…

…the mechanism is a **mutable global static on `MetricTransformer`**, which is a pure wire encoder today. That global bleeds across meter providers and across tests (the same reason `OTelFactory` state gets reset so carefully everywhere else in this codebase), and it makes the transformer's output depend on hidden state.

The good news: the right home already exists in this repo. The unconditional `exemplars.add(exemplar)` calls in `lib/src/metrics/storage/gauge_storage.dart` and `histogram_storage.dart` are the spec's "offer" point — gate those on the filter (with `trace_based` checking the current span context) and the transformer stays a pure encoder. That's how the Java/Go/Python SDKs structure it too. The env parsing and `MetricsSdkConfig` plumbing in this PR are right and worth keeping as-is; happy to split so interval/timeout land now and the storage-layer filter follows as its own PR if you'd like a checkpoint on the design before writing it.

lib/src/metrics/export/otlp/metric_transformer.dart
```diff
@@ -223,4 +215,26 @@ class MetricTransformer {
...
+      switch (_exemplarFilter) {
+        case MetricsExemplarFilter.alwaysOn:
+          return true;
+        case MetricsExemplarFilter.alwaysOff:
+          return false;
+        case MetricsExemplarFilter.traceBased:
+          return exemplar.traceId != null || exemplar.spanId != null;
+      }
+    }).map((exemplar) {
+      return proto.Exemplar(
```
michaelbushe, line 234: Three wire-correctness issues in the exemplar encoding itself, independent of where the filter lives:

1. **The trace context is dropped on the floor**: `proto.Exemplar` is built with only `timeUnixNano` and `asDouble` — no `traceId`/`spanId` bytes, no `filteredAttributes`. An exemplar without its trace/span ids can't link a metric point to the trace that produced it, which is the entire point of `trace_based` exemplars. (`TraceId.bytes`/`SpanId.bytes` give you the 16/8-byte values the proto wants.)
2. **`asDouble` unconditionally coerces int measurements**: proto `Exemplar` is a oneof — integer measurements should encode as `asInt`, doubles as `asDouble`. Always using `asDouble` changes the wire type for counter exemplars.
3. **The `trace_based` predicate** at line 231: `traceId != null || spanId != null` accepts half-formed context. Spec semantics are "measurements with a *sampled* trace context" — at minimum this should be `&&` plus a validity check (`traceId.isValid`), or exemplars from unsampled/invalid contexts leak through.

Points 1–2 matter even if the filter moves layers, so they're worth fixing here regardless of how the architecture discussion lands.

lib/src/metrics/export/metric_config.dart
```diff
@@ -66,7 +69,8 @@ class MetricsConfiguration {
 
     metricReader ??= PeriodicExportingMetricReader(
       metricExporter,
-      interval: const Duration(seconds: 15),
+      interval: metricsSdkConfig.exportInterval,
```
michaelbushe, line 72: Spec-correct default (`OTEL_METRIC_EXPORT_INTERVAL` defaults to 60000 ms) — but note this changes the SDK's out-of-the-box behavior from the previous hardcoded 15 s to 60 s, so it must be called out in the CHANGELOG under `## [1.1.0-beta.10-wip]` with the "set `OTEL_METRIC_EXPORT_INTERVAL=15000` to restore the old cadence" escape hatch. There's precedent wording in the beta.8 notes for the BSP 1 s → 5 s change. (The PR currently has no CHANGELOG entry at all — please add one covering all three new vars.)

lib/src/metrics/metric_reader.dart
```diff
@@ -75,6 +75,12 @@ class PeriodicExportingMetricReader extends MetricReader {
     _startTimer();
   }
 
+  /// How often metrics are exported.
+  Duration get interval => _interval;
```
michaelbushe, line 79: Exposing `interval`/`timeout` as getters is the right call — it's what makes the pipeline shape assertable in tests instead of only observable through timing, same pattern the exporter-selection tests rely on elsewhere. 👍

michaelbushe (changes requested):

lib/src/metrics/export/metric_config.dart
michaelbushe, line 1: `OTEL_METRICS_EXEMPLAR_FILTER` is parsed but never applied.

If you want this to just be arg passing, that's ok , we can do another PR for the application before the next release.

`configureMeterProvider` reads `MetricsSdkConfig.fromEnvironment()` and wires `exportInterval` and `exportTimeout` into the reader, but never touches `metricsSdkConfig.exemplarFilter`. Across the whole diff, the only callers of `MetricTransformer.setExemplarFilter()` are the two test files — so setting the variable has no runtime effect, while the CHANGELOG says it "configures the exemplar filtering policy."

The fix is one line at this seam, next to the interval/timeout wiring. I've pushed two failing tests (`test/unit/metrics/exemplar_filter_wiring_test.dart`) that pin it end-to-end: they use an exemplar with no trace context as the witness, since it survives only under `always_on` and disappears only under `always_off` — so they can pass only when the parsed value genuinely reaches the transformer.

Worth knowing while you're here: nothing in the SDK ever calls `addExemplar()` from a measurement path, so even once wired the filter gates a set that is always empty in practice. The reservoir/sampling work is #154, which is still open and unclaimed. So either wire it and reword the CHANGELOG bullet as groundwork ahead of #154, or drop that bullet and ship the two variables that fully work today — both are fine, but the current combination claims something the code doesn't do.

lib/src/metrics/export/otlp/metric_transformer.dart
michaelbushe, line 1: Design point that pairs with the wiring fix: `_exemplarFilter` is a mutable **static**, so the filter is process-global. Two MeterProviders in one process can't have different policies, and it leaks between tests — the tests you added have to keep re-setting it, and the ones I pushed have to reset it in `setUp`/`tearDown` for the same reason.

Since the value already flows through `MetricsSdkConfig`, it would sit more naturally on the reader/exporter config path alongside `interval` and `timeout`, and get passed into the transform call rather than read from static state. That also makes the wiring testable without global reset dances.

The filtering logic itself (the `alwaysOn` / `alwaysOff` / `traceBased` switch, with `traceBased` requiring a valid trace *and* span id) is correct — this is only about where the setting lives.

lib/src/metrics/export/metrics_sdk_config.dart
michaelbushe, line 1: Two notes on `fromEnvironment`, one blocking and one informational.

**Blocking: a zero interval slips through.** `getPositiveIntEnv(otelMetricExportInterval, minInclusive: 0)` accepts `0`, and the guard here only rejects `< 0`, so `OTEL_METRIC_EXPORT_INTERVAL=0` becomes `Duration.zero` and lands in `Timer.periodic(Duration.zero, ...)` (`metric_reader.dart:81`). That fires on every event-loop turn — a CPU spin that also floods the exporter. The spec's `Duration` type only mandates non-negative, so `0` isn't strictly illegal, but Java rejects non-positive intervals and Go falls back to its default; an operator typo shouldn't be able to hot-loop the process. `minInclusive: 1`, or coercing `0` to the default with a warning, closes it. Failing test pushed in `metrics_sdk_config_test.dart` ("a zero export interval falls back to the default").

**Informational: the negative guards are unreachable from the environment.** `getPositiveIntEnv` already rejects negative values *and emits its own warning* (`OTelEnv: Illegal value for $name: $value is below minimum ...`), so `env.exportInterval` / `env.exportTimeout` can never arrive negative here — these `< 0` branches are only reachable through the test-only `overrides` parameter, which your tests do exercise. No compliance gap: the spec's MUST-warn-on-negative is satisfied one layer up. Just worth a comment (or removal) so the next reader doesn't assume env values reach it.

**And a compliment on the part that's easy to get wrong:** treating `OTEL_METRIC_EXPORT_TIMEOUT=0` as "no limit" via a very large duration is exactly right. `configuration/common.md` separates the `Timeout` type from `Duration` for precisely this reason — *"Implementations SHOULD interpret timeout zero values ... as no limit (i.e. infinite). In practice, implementations MAY treat no limit as 'a very long time' and substitute a very large duration."* Nice catch.

michaelbushe (changes requested):

lib/src/metrics/export/metric_config.dart
harshitt13: The env var parsing and plumbing down to the exporters is fully wired up now, and the static state leak in `MetricTransformer` has been completely removed. I also updated the `CHANGELOG.md` to clarify that this is groundwork ahead of #154, and that the filtering is currently inactive since `addExemplar()` isn't used yet. This keeps the environment variable work cleanly scoped to this PR.

michaelbushe:
Nice job, @harshitt13 .  I liked that you deleted the static entirely and threaded the filter through OtlpGrpcMetricExporterConfig.exemplarFilter.  I added in some more tests.

PR #74, opened 2026-07-13
feat: parse general attribute limits in OTelEnv

### LINKED ISSUE #73: [FEATURE] Implement General Attribute Limits
Ref #20

Implementation of the General Attribute Limits. 

**Variables to implement:**
* `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT`
* `OTEL_ATTRIBUTE_COUNT_LIMIT` (default: 128)

**Implementation Notes:**
* These limits apply globally as a baseline to all telemetry signals (Traces, Logs, Metrics).
* Implement truncation logic for attribute string values that exceed `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT`.
* Implement dropping logic for excess attributes that exceed `OTEL_ATTRIBUTE_COUNT_LIMIT`.
* Create a centralized utility to handle attribute truncation and dropping cleanly across the SDK.
* Log warnings when limits are exceeded.
* Export metrics (e.g., `droppedAttributesCount`) to OTLP payload when attributes are dropped.

harshitt13:
Closes #73 

Used Opus 4.6

### DIFF
--- a/lib/src/environment/attribute_limits_config.dart
+++ b/lib/src/environment/attribute_limits_config.dart
@@ -1,0 +1,56 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+/// Configuration for attribute limits read from environment variables.
+///
+/// Used by both general attribute limits (`OTEL_ATTRIBUTE_*`) and
+/// signal-specific limits (e.g., `OTEL_LOGRECORD_ATTRIBUTE_*`).
+///
+/// Per the OpenTelemetry specification:
+/// - Values exceeding the length limit should be truncated.
+/// - Attributes exceeding the count limit should be dropped.
+/// - Warnings should be logged when limits are exceeded.
+///
+/// See: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#attribute-limits
+class AttributeLimitsConfig {
+  /// Maximum length of attribute values.
+  ///
+  /// When set, attribute values longer than this limit should be truncated.
+  /// `null` means unlimited (no limit set).
+  final int? attributeValueLengthLimit;
+
+  /// Maximum number of attributes allowed per telemetry item.
+  ///
+  /// When set, attributes beyond this count should be dropped.
+  /// `null` means not explicitly configured via environment variable.
+  /// The spec default is 128 when not set.
+  final int? attributeCountLimit;
+
+  /// Creates a new [AttributeLimitsConfig].
+  const AttributeLimitsConfig({
+    this.attributeValueLengthLimit,
+    this.attributeCountLimit,
+  });
+
+  /// Whether no limits were configured from environment variables.
+  bool get isEmpty =>
+      attributeValueLengthLimit == null && attributeCountLimit == null;
+
+  /// Whether any limits were configured from environment variables.
+  bool get isNotEmpty => !isEmpty;
+
+  /// Converts this config to a JSON-compatible map.
+  ///
+  /// Only includes fields that are non-null, matching the previous
+  /// `Map<String, dynamic>` behavior for backward compatibility.
+  Map<String, dynamic> toJson() {
+    final map = <String, dynamic>{};
+    if (attributeValueLengthLimit != null) {
+      map['attributeValueLengthLimit'] = attributeValueLengthLimit;
+    }
+    if (attributeCountLimit != null) {
+      map['attributeCountLimit'] = attributeCountLimit;
+    }
+    return map;
+  }
+}
--- a/lib/src/environment/otel_env.dart
+++ b/lib/src/environment/otel_env.dart
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import 'package:dartastic_opentelemetry_api/dartastic_opentelemetry_api.dart';
+import 'attribute_limits_config.dart';
 import 'env_constants.dart';
 import 'environment_service.dart';
 
@@ -443,12 +444,8 @@ class OTelEnv {
   /// logs) unless overridden by signal-specific limits (e.g., span or
   /// log record attribute limits).
   ///
-  /// Returns a map containing the general attribute limits.
-  /// Keys returned:
-  /// - 'attributeValueLengthLimit': int for max attribute value length
-  ///   (no default — unlimited when not set)
-  /// - 'attributeCountLimit': int for max number of attributes per item
-  ///   (spec default: 128)
+  /// Returns an [AttributeLimitsConfig] containing the general attribute
+  /// limits. Fields that are not set via environment variables will be `null`.
   ///
   /// Per the OpenTelemetry specification:
   /// - Values exceeding the length limit should be truncated.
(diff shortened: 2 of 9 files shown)


lib/src/environment/otel_env.dart
```diff
@@ -437,6 +437,59 @@ class OTelEnv {
...
+        if (OTelLog.isDebug()) {
+          OTelLog.debug(
+            'OTelEnv: General attribute count limit set to $limit',
+          );
+        }
+      }
+    }
+
+    return config;
+  }
```
yuzurihaaa: Would it be better here if this is in class instead of map? Better typing too.

harshitt13: Refactored both `getAttributeLimits()` and `getLogRecordLimits()` to return a typed `AttributeLimitsConfig` class instead of Map.

yuzurihaaa (approved):

lib/src/environment/otel_env.dart
```diff
@@ -437,21 +438,73 @@ class OTelEnv {
...
+  /// Returns an [AttributeLimitsConfig] containing the general attribute
+  /// limits. Fields that are not set via environment variables will be `null`.
+  ///
+  /// Per the OpenTelemetry specification:
+  /// - Values exceeding the length limit should be truncated.
+  /// - Attributes exceeding the count limit should be dropped.
+  /// - Warnings should be logged when limits are exceeded.
+  ///
+  /// See: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#attribute-limits
+  static AttributeLimitsConfig getAttributeLimits() {
```
michaelbushe, line 456: The parsing here is solid, but nothing in `lib/` consumes it — the only callers of `getAttributeLimits()` are the tests. The spec behavior this enables (truncate over-length values, drop excess attributes, warn when limits are exceeded) isn't implemented, and the fallback chain the spec defines (signal-specific limits override the general ones; when `OTEL_LOGRECORD_ATTRIBUTE_*` is unset, `OTEL_ATTRIBUTE_*` applies) isn't wired into `getLogRecordLimits()` below either.

That's fine as a first slice! But then the PR should say so: please change "Closes #73" to "Part of #73" and note the follow-ups (enforcement at the attribute-creation/record boundaries, plus the general→signal fallback). Otherwise #73 auto-closes with the env vars parsed but having no observable effect, which is the kind of gap that surfaces as a confusing bug report later. The fallback wiring would also make a nice, self-contained next PR.

lib/src/environment/otel_env.dart
```diff
@@ -437,21 +438,73 @@ class OTelEnv {
...
+    int? parsedValueLengthLimit;
+    int? parsedCountLimit;
+
+    // Get attribute value length limit
+    final valueLengthLimit = _getEnv(otelAttributeValueLengthLimit);
+    if (valueLengthLimit != null) {
+      final limit = int.tryParse(valueLengthLimit);
```
michaelbushe, line 463: `int.tryParse` happily accepts `-5`, and a negative limit is meaningless — please reject non-negative violations, and warn (not debug, and not silence) when a set value is unusable, per the spec's guidance and consistent with where #66 landed. If you fold the parsing into a shared helper (see the config-class comment), the range check and the warn live in exactly one place for both the general and logrecord variants.

lib/src/environment/attribute_limits_config.dart
```diff
@@ -0,0 +1,56 @@
...
+/// Used by both general attribute limits (`OTEL_ATTRIBUTE_*`) and
+/// signal-specific limits (e.g., `OTEL_LOGRECORD_ATTRIBUTE_*`).
+///
+/// Per the OpenTelemetry specification:
+/// - Values exceeding the length limit should be truncated.
+/// - Attributes exceeding the count limit should be dropped.
+/// - Warnings should be logged when limits are exceeded.
+///
+/// See: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#attribute-limits
+class AttributeLimitsConfig {
```
michaelbushe, line 15: This is the pattern, nicely done — typed config class, CNCF header, barrel export, all right on the first pass. One structural upgrade: `getAttributeLimits()` and `getLogRecordLimits()` are now line-for-line copies differing only in which env vars they read. A single parameterized factory here collapses both:

```dart
factory AttributeLimitsConfig.fromEnvironment({
  required String valueLengthVar,
  required String countVar,
}) { ... }
```

with `OTelEnv` (or callers) passing `otelAttributeValueLengthLimit`/`otelLogrecordAttributeValueLengthLimit` etc. Parsing, range validation, and the warn then exist exactly once, and the future span/event/link limit vars (`OTEL_SPAN_ATTRIBUTE_*` and friends, already in `env_constants.dart`) become one-liners.

lib/src/environment/attribute_limits_config.dart
```diff
@@ -0,0 +1,56 @@
...
+      attributeValueLengthLimit == null && attributeCountLimit == null;
+
+  /// Whether any limits were configured from environment variables.
+  bool get isNotEmpty => !isEmpty;
+
+  /// Converts this config to a JSON-compatible map.
+  ///
+  /// Only includes fields that are non-null, matching the previous
+  /// `Map<String, dynamic>` behavior for backward compatibility.
+  Map<String, dynamic> toJson() {
```
michaelbushe, line 46: `toJson()` has no production consumer — only the test helpers use it, and they can build their own maps. "Matching the previous Map behavior for backward compatibility" isn't needed here since the map shape was never a published API. Pre-1.0, the cheapest API surface to maintain is the one you didn't ship — suggest dropping it (or moving it into the test helper if the subprocess tests want JSON).

lib/src/environment/otel_env.dart
```diff
@@ -437,21 +438,73 @@ class OTelEnv {
...
   ///
-  /// Returns a map containing the log record attribute limits.
-  /// Keys returned:
-  /// - 'attributeValueLengthLimit': int for max attribute value length
-  /// - 'attributeCountLimit': int for max number of attributes
-  static Map<String, dynamic> getLogRecordLimits() {
-    final config = <String, dynamic>{};
+  /// Returns an [AttributeLimitsConfig] containing the log record attribute
+  /// limits. Fields that are not set via environment variables will be `null`.
+  static AttributeLimitsConfig getLogRecordLimits() {
```
michaelbushe, line 498: Rebase heads-up: since #84 merged, main has in-process tests (`test/unit/environment/otel_env_inprocess_test.dart`) that assert `getLogRecordLimits()` returns the *map* shape — changing the return type to `AttributeLimitsConfig` will break them until they're updated to the new fields (a mechanical fix, and honestly an improvement to those assertions). Main also now has `EnvironmentService.testOverrides`, which lets you test all of this in-process without the subprocess helpers — worth a look before adding more helper scripts.

And one more for the checklist: no CHANGELOG entry yet — please add one under `## [1.1.0-beta.10-wip]` for the new `OTEL_ATTRIBUTE_*` parsing (noting enforcement is follow-up work).

michaelbushe (changes requested):

lib/src/environment/otel_env.dart
```diff
@@ -652,29 +664,74 @@ class OTelEnv {
     );
   }
 
+  static ({int? attributeValueLengthLimit, int? attributeCountLimit})
```
michaelbushe: For consistency and readability, this should return AttributeLimitsEnvironmentValues.

michaelbushe (changes requested):
Just one change - return AttributeLimitsEnvironmentValues, and I pushed some log tests. Since logs are output that can be relied on and behavior that should not change, we should test for them.

michaelbushe:
Thanks @harshitt13 !!!

Issue #80, opened 2026-07-18
[FEATURE] Zipkin span exporter (OTEL_TRACES_EXPORTER=zipkin)


michaelbushe:
**Target: 1.1.0-beta.11 — up for grabs.** Good second contribution after the env-var wave: well-specified, self-contained, and the wiring seams already exist.

## The spec

- sdk-environment-variables.md, "Exporter Selection": `zipkin` is a known value for `OTEL_TRACES_EXPORTER` — *"Zipkin (Defaults to protobuf format)"*; we'll implement the JSON v2 API (`OTEL_EXPORTER_ZIPKIN_PROTOCOL` is reserved/unspecified as of spec 1.0, so JSON v2 over HTTP is the conventional choice — it's what most SIGs ship).
- Zipkin exporter env vars (already in `env_constants.dart:597-610`, currently consumed by nothing): `OTEL_EXPORTER_ZIPKIN_ENDPOINT` (default `http://localhost:9411/api/v2/spans`), `OTEL_EXPORTER_ZIPKIN_TIMEOUT` (default 10000 ms).
- The span mapping is fully specified in `specification/trace/sdk_exporters/zipkin.md`: hex `traceId`/`id`/`parentId`; `kind` mapping (INTERNAL has no Zipkin kind — omit); `timestamp`/`duration` in microseconds; `localEndpoint.serviceName` from `service.name`; events → `annotations`; attributes → string-valued `tags`; status → `otel.status_code` tag plus `error` tag on ERROR (the `error` tag carries the status description, empty string if none); `InstrumentationScope` → `otel.scope.name`/`otel.scope.version` tags; `remoteEndpoint` derivation for CLIENT/PRODUCER per the spec's attribute precedence list.

## Scope

1. `ZipkinSpanExporter implements SpanExporter` — HTTP POST of the JSON v2 array, honoring endpoint/timeout env vars with config-object overrides, `package:http` like the OTLP HTTP exporters; shutdown/forceFlush semantics consistent with `OtlpHttpSpanExporter` (retry can be a follow-up).
2. `ZipkinSpanTransformer` — the mapping above, as a pure function over `Span`s (mirrors `OtlpSpanTransformer`'s layout).
[...]

PR #81, opened 2026-07-18
test: pin the Zipkin exporter gap (#80 — up for grabs)

michaelbushe:
Pins the missing-Zipkin-exporter gap in the house TDD style (same shape as #41/#43): one spec-quoting contract, skipped behind #80 so CI stays green until the implementation lands.

**#80 has the full scope, mapping references, and acceptance criteria — comment there to claim it.** The seams are ready: the `OTEL_EXPORTER_ZIPKIN_*` constants already exist, and after #79 the `OTEL_TRACES_EXPORTER` switch has an exact spot where `'zipkin'` currently warns-and-ignores.

Targeted 1.1.0-beta.11. Whoever takes it up: replace this placeholder with the real contracts as you go — transformer mapping tests, a `spec_defaults_test.dart` pipeline contract, and an HTTP capture test (use `await expectLater`, no fixed sleeps — see #77 for why).

### DIFF
--- a/test/unit/trace/export/zipkin/zipkin_exporter_gap_test.dart
+++ b/test/unit/trace/export/zipkin/zipkin_exporter_gap_test.dart
@@ -1,0 +1,26 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:test/test.dart';
+
+// Spec-compliance gap for sdk-environment-variables.md, "Exporter
+// Selection" and trace/sdk_exporters/zipkin.md:
+//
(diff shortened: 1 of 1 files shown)


michaelbushe:
Withdrawn — the Zipkin exporter spec is **Deprecated** with removal from the OpenTelemetry specification scheduled December 2026, and it explicitly states "Implementing a Zipkin exporter is not required for new SDKs" (trace/sdk_exporters/zipkin.md). The spec-correct posture is what #79 already does: warn-and-ignore, with the collector's zipkin exporter as the migration path for anyone who needs the format.

Issue #96, opened 2026-08-08
[FEATURE] OTLP Headers log REDACTED list


michaelbushe:
In `otel_env.dart` the `authorization` header value gets REDACTED when output but OTelLog.debug():

```dart      final parsedHeaders = _parseHeaders(headers);
      if (OTelLog.isDebug()) {
        OTelLog.debug('OTelEnv: Parsed ${parsedHeaders.length} header(s)');
        parsedHeaders.forEach((key, value) {
          if (key.toLowerCase() == 'authorization') {
            OTelLog.debug('  $key: [REDACTED - length: ${value.length}]');
          } else {
            OTelLog.debug('  $key: $value');
          }
        });
      }
      config['headers'] = parsedHeaders;
...
```
It's possible that SDK users would want other values redacted as well.

This would need a `otlpHeaderRedacted List<String>` in `initialize()`

arpitjain099:
Worth knowing before a redaction list lands: the redaction quoted here is bypassed today.

A few lines above it, `getOtlpConfig` logs the unparsed env var:

```dart
OTelLog.debug('OTelEnv: Parsing $signal headers from env: $headers');
```

`$headers` is the raw `OTEL_EXPORTER_OTLP_HEADERS` string, so it already contains the `authorization` value that the loop below then carefully redacts. With debug logging on, the token reached the log either way. I opened #100 for that one line, with a test that captures `OTelLog` output and asserts the token never appears while the `REDACTED` marker and non-secret headers still do.

That does not close this issue, and I did not want to fold the two together. A few design questions on the list itself, since they change the implementation:

- Denylist or allowlist? Header names that carry secrets are open ended (`x-api-key`, `api-key`, `x-honeycomb-team`, `dd-api-key`, and whatever a given backend invents). A denylist fails open on anything nobody thought of; an allowlist of names safe to print fails closed. Given this is credential material, failing closed seems the safer default even though it is noisier.
- Matching: exact lowercase name only, or prefix and substring so `x-vendor-token` is covered without being enumerated?
- Configuration surface: `initialize()` only, or also an env var? The headers themselves arrive by env var, so someone configuring OTLP entirely through the environment cannot reach an `initialize()` parameter.

Happy to implement whichever shape you prefer, or to leave it with you if you already have one in mind.

michaelbushe:
These are all good point, nice catches.  
- Yes, an allowlist failing closed is preferred. It's also more efficient. 
[...]

PR #100, opened 2026-08-10
Stop logging the raw OTEL_EXPORTER_OTLP_HEADERS value

arpitjain099:
## What this fixes

`OTelEnv.getOtlpConfig` logs the raw `OTEL_EXPORTER_OTLP_HEADERS` value at debug level:

```dart
OTelLog.debug('OTelEnv: Parsing $signal headers from env: $headers');
```

Five lines below it, the per-header loop is careful to keep the credential out of the log:

```dart
if (key.toLowerCase() == 'authorization') {
  OTelLog.debug('  $key: [REDACTED - length: ${value.length}]');
}
```

The raw string logged above contains that same value, so the redaction never had any effect. With debug logging on, `authorization=Bearer <token>` reached the log regardless. This is the shape where a guard exists and something upstream of it makes the guard moot, so it reads as protected while it is not.

The fix keeps the message and drops the value. The loop underneath still reports every header, still redacts Authorization, and still prints non-secret values, so nothing diagnostic is lost.

## Relationship to #96

#96 asks for a configurable redaction list for OTLP headers. This is not that, and it does not close it. It is the narrower point that the redaction already in the file is bypassed today, which seemed worth fixing on its own rather than folding into the broader design. Happy to help with #96 separately if the list approach is the direction you want.

## Test

`test/unit/environment/helpers/check_header_logging.dart` captures `OTelLog` output in a subprocess with the env var set, following the existing `runWithEnv` pattern in this suite, and prints it as JSON. The test asserts three things: the token never appears, the `REDACTED` marker still does, and the non-secret header value still does, so a regression that simply silenced all header logging would fail it too.

Without the change the test fails on the first assertion:

```
the Authorization value must never reach the log
test/unit/environment/env_coverage_test.dart 517:7
00:02 +0 -1: Some tests failed.
```

With it, `dart test test/unit/environment/` is green at 282 passed, and `dart analyze` reports no issues.

## Note on tooling

I used AI assistance while investigating this. I read the file, made the change and ran the tests above myself, including the failing run, so the output quoted here is from runs I did rather than reasoning about.

### DIFF
--- a/lib/src/environment/otel_env.dart
+++ b/lib/src/environment/otel_env.dart
@@ -161,7 +161,9 @@ class OTelEnv {
     }
     if (headers != null) {
       if (OTelLog.isDebug()) {
-        OTelLog.debug('OTelEnv: Parsing $signal headers from env: $headers');
+        // The raw value is not logged: it carries the Authorization header,
+        // which the per-header loop below is careful to redact.
+        OTelLog.debug('OTelEnv: Parsing $signal headers from env');
       }
       final parsedHeaders = _parseHeaders(headers);
       if (OTelLog.isDebug()) {
--- a/test/unit/environment/env_coverage_test.dart
+++ b/test/unit/environment/env_coverage_test.dart
@@ -499,6 +499,30 @@ void main() {
       expect(result.containsKey('key'), isFalse);
     });
 
+    test('does not log the raw header string, which carries the token',
+        () async {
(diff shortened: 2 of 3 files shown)


michaelbushe (approved):
Very good and timely, thank you!

michaelbushe:
This was merged into #101 and released in beta.13, announced in a GSA.

PR #101, opened 2026-08-12
Redact OTLP header values in debug logs behind an allowlist

### LINKED ISSUE #96: [FEATURE] OTLP Headers log REDACTED list

arpitjain099:
Closes #96.

Built to the scope in your comment. Stacked on #100, so the first commit here is that one-line fix; the diff to review is the second commit. Happy to rebase once #100 lands.

## What it does

A shared helper in `lib/src/util/header_redaction.dart`, applied at every site that logs headers. There were five, all carrying the same `authorization`-only check:

- `environment/otel_env.dart` (env parsing)
- `trace/export/otlp/http/otlp_http_span_exporter.dart`, config and request paths
- `logs/export/otlp/http/otlp_http_log_record_exporter.dart`, config and request paths

The gRPC exporters do not log header values, so nothing to do there.

Following your points:

- **Allowlist, fails closed.** Default redacts every value.
- **Exact match only.** Lowercased once at config time into a `Set<String>`, and the header lowercased at compare time. A test pins the reason prefixes are out: an `x-trace` prefix would opt in `x-api-key`.
- **Env var and `initialize()` param.** `DAR_OTLP_HEADER_LOG_ALLOWLIST` (comma separated, trimmed, empties dropped, deduplicated) and `otlpHeaderLogAllowlist:`. The parameter replaces the env var, no merging.
- **Always denied.** `authorization` and `proxy-authorization` are stripped from the allowlist at config time, so listing them cannot opt them in.
- **No length in the placeholder.** Just `[REDACTED]`.
- **Names and count still logged.**

## What the output looks like

Default:

```
OTelEnv: Parsed 3 header(s)
  authorization: [REDACTED]
  x-api-key: [REDACTED]
  x-trace-id: [REDACTED]
```

With `DAR_OTLP_HEADER_LOG_ALLOWLIST=x-trace-id`, only `x-trace-id` shows its value.

## Behaviour change

As you said, this is a security fix rather than an API break: non-`authorization` values were previously logged in the clear and now are not. It changed one existing assertion, in the test I added in #100, which asserted a non-secret value was still printed. It now asserts the name is still printed instead, which is the property that actually matters for debuggability.

## Docs

README gets a "Header Value Redaction in Debug Logs" section covering the default, why it is an allowlist and not a denylist, that it opts in values and never hides names, the exact-match rule, the always-denied pair, and why the variable is `DAR_` rather than `OTEL_` prefixed. The env var table gets a row.

## Tests

`test/unit/util/header_redaction_test.dart` covers the helper directly: default, allowlisting, case insensitivity on both sides, exact-vs-prefix, the always-denied pair surviving an attempt to allowlist them, replacement rather than accumulation, and the parser's trim/empty/dedupe behaviour.

`env_coverage_test.dart` asserts the actual log output through the subprocess helper, as you asked: everything redacted by default, only the allowlisted header showing a value, `authorization` still redacted when the env var names it, and the parameter beating the env var. I extended `runWithEnv` to pass script arguments and to clear `DAR_` vars alongside `OTEL_` ones.

`dart analyze` clean, `dart format` clean, `dart test` is 2017 passing. The 4 failures in `test/unit/context/context_propagation_test.dart` are pre-existing; I confirmed they fail the same way on a clean `main` checkout.

### DIFF
--- a/lib/dartastic_opentelemetry.dart
+++ b/lib/dartastic_opentelemetry.dart
@@ -94,3 +94,4 @@ export 'src/trace/span_processor.dart';
 export 'src/trace/tracer.dart' hide SDKTracerCreate;
 export 'src/trace/tracer_provider.dart' hide SDKTracerProviderCreate;
 export 'src/trace/w3c_trace_context_propagator.dart';
+export 'src/util/header_redaction.dart';
--- a/lib/dartastic_opentelemetry.dart
+++ b/lib/dartastic_opentelemetry.dart
@@ -94,4 +94,3 @@ export 'src/trace/span_processor.dart';
 export 'src/trace/tracer.dart' hide SDKTracerCreate;
 export 'src/trace/tracer_provider.dart' hide SDKTracerProviderCreate;
 export 'src/trace/w3c_trace_context_propagator.dart';
-export 'src/util/header_redaction.dart';
--- a/lib/src/environment/env_constants.dart
+++ b/lib/src/environment/env_constants.dart
@@ -124,6 +124,18 @@ const String otelDartLogExport = 'OTEL_DART_LOG_EXPORT';
 ///
 /// Type: String (URL)
 /// Default: "http://localhost:4318" (HTTP) or "http://localhost:4317" (gRPC)
+/// Comma separated list of OTLP header names whose values may be written to the
+/// debug log.
+///
+/// Not an `OTEL_` name because it is not in the OpenTelemetry specification.
+/// Names are matched exactly, case insensitively. `authorization` and
+/// `proxy-authorization` are never logged even if listed. Any header not listed
+/// has its value replaced with `[REDACTED]`; the name and the header count are
+/// still logged.
+///
+/// Example: `DAR_OTLP_HEADER_LOG_ALLOWLIST=x-trace-id,x-tenant`
+const String darOtlpHeaderLogAllowlist = 'DAR_OTLP_HEADER_LOG_ALLOWLIST';
+
 const String otelExporterOtlpEndpoint = 'OTEL_EXPORTER_OTLP_ENDPOINT';
 
 /// Transport protocol for the OTLP exporter.
--- a/lib/src/environment/env_constants.dart
+++ b/lib/src/environment/env_constants.dart
@@ -110,6 +110,14 @@ const String otelDartLogSpans = 'OTEL_DART_LOG_SPANS';
 /// Default: false
 const String otelDartLogExport = 'OTEL_DART_LOG_EXPORT';
 
+/// Comma separated OTLP header names whose values may appear in the debug log
+/// (Dart-specific). Names are matched case insensitively, and `authorization`
+/// and `proxy-authorization` are redacted even when listed.
+///
+/// Type: String (comma-separated header names)
+/// Default: none, every header value is redacted
+const String otelDartHeaderLogAllowlist = 'OTEL_DART_HEADER_LOG_ALLOWLIST';
+
 // =============================================================================
 // General OTLP Exporter Configuration
 // =============================================================================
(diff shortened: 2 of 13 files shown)


lib/src/environment/env_constants.dart
```diff
@@ -124,6 +124,18 @@ const String otelDartLogExport = 'OTEL_DART_LOG_EXPORT';
...
+/// debug log.
+///
+/// Not an `OTEL_` name because it is not in the OpenTelemetry specification.
+/// Names are matched exactly, case insensitively. `authorization` and
+/// `proxy-authorization` are never logged even if listed. Any header not listed
+/// has its value replaced with `[REDACTED]`; the name and the header count are
+/// still logged.
+///
+/// Example: `DAR_OTLP_HEADER_LOG_ALLOWLIST=x-trace-id,x-tenant`
+const String darOtlpHeaderLogAllowlist = 'DAR_OTLP_HEADER_LOG_ALLOWLIST';
```
michaelbushe: DAR_ is not correct but my earlier guidance of DARTASTIC_ is also not correct.  Dartastic.io's vendor extensions use DARTASTIC_, I should have been more careful.  This is intended to be the language SDK so we should use the language name DART_

If you look at the surrounding context you see: OTEL_DART_....  This should also follow the OTEL_{LANGUAGE}_{FEATURE} pattern, and thus be OTEL_DART_HEADER_LOG_ALLOWLIST.  The variable should be renamed likewise.

Also the placement of the variable is below the doc for otelExpoerterOtlpEndpoint.  The dartdoc will then have exporter doc and your the doc for the new field and the exporter doc have none.  Again, human eyes should avoid such problems.

lib/dartastic_opentelemetry.dart
```diff
@@ -94,3 +94,4 @@ export 'src/trace/span_processor.dart';
 export 'src/trace/tracer.dart' hide SDKTracerCreate;
 export 'src/trace/tracer_provider.dart' hide SDKTracerProviderCreate;
 export 'src/trace/w3c_trace_context_propagator.dart';
+export 'src/util/header_redaction.dart';
```
michaelbushe: This is internal and should not be exported.

lib/src/environment/env_constants.dart
arpitjain099: Renamed to `OTEL_DART_HEADER_LOG_ALLOWLIST`, and the Dart constant to `otelDartHeaderLogAllowlist`. It now sits in the Dart-specific section next to `otelDartLogMetrics`, `otelDartLogSpans` and `otelDartLogExport`, which is where it should have gone in the first place.

The placement was worse than misplaced doc. I had wedged the constant between the dartdoc for `otelExporterOtlpEndpoint` and its declaration, so the endpoint's own documentation was attached to my constant and the endpoint had none at all. Both are correct now.

While fixing the name I also dropped the `DAR_` prefix clearing I had added to `runWithEnv` in the test helper, since the variable starts with `OTEL_` again, and the README note explaining the non-`OTEL_` prefix, which no longer applies.

lib/dartastic_opentelemetry.dart
arpitjain099: Export removed. The redaction helper is an implementation detail of the logging paths, not a knob callers reach for, so the public surface stays the `otlpHeaderLogAllowlist` parameter and the environment variable. Its test imports `package:dartastic_opentelemetry/src/util/header_redaction.dart` directly, matching the 29 other test files that already import from `src/`.

arpitjain099:
Fair hit on the review, both of those were mine to catch. The dartdoc one especially: I inserted a constant between an existing doc comment and its declaration and did not look at what that did to the field above it.

On the comments, I went back over the whole diff and cut what was narrating rather than telling you something. The `expect` calls in `env_coverage_test.dart` had eight `reason:` arguments; that file has zero, so they were mine and not the file's, and they are gone. The comment above `OTelEnv.applyHeaderLogAllowlist` in `otel.dart` was restating the dartdoc on the method it calls, so it is one line now saying the only thing that is not obvious, which is that it has to run first. Same treatment for the library doc on `header_redaction.dart` and the two subprocess helper headers, which were three times the length of the helpers next to them.

`dart analyze` clean, `dart format` clean, 298 tests pass across `test/unit/util` and `test/unit/environment`. `test/unit/context/context_propagation_test.dart` has 4 failures, but they fail the same way on an unmodified `main`, so they are not from this branch.

michaelbushe:
@arpitjain099 Please accept credit on the advisory.
https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/security/advisories/GHSA-4rh6-c2v5-374w

PR #104, opened 2026-08-13
fix: OTEL_RESOURCE_ATTRIBUTES no longer overrides higher-precedence service.name

### LINKED ISSUE #103: OTEL_RESOURCE_ATTRIBUTES service.name overrides both OTEL_SERVICE_NAME and explicit config (spec violation)
Originally filed as "tests assume a clean environment" (split from #98 / PR #99). Investigation shows the premise was wrong: **the tests are right and the library is wrong.**

## The bug

`EnvVarResourceDetector` merges `OTEL_RESOURCE_ATTRIBUTES` over the already-resolved resource in `OTel.initialize`, so `service.name` from that variable clobbers values that must outrank it.

Probed directly:

| Case | Result | Expected |
|---|---|---|
| `OTelEnv.getServiceConfig()` with `OTEL_SERVICE_NAME` + `OTEL_RESOURCE_ATTRIBUTES` | `{serviceName: from-service-name}` | ✅ correct |
| `OTel.initialize()`, same environment | `service.name=from-resource-attrs` | ❌ should be `from-service-name` |
| `OTel.initialize(serviceName: 'explicit-service')` + `OTEL_RESOURCE_ATTRIBUTES` | `from-resource-attrs` | ❌ should be `explicit-service` |

Two spec violations from one root cause:

1. **`OTEL_SERVICE_NAME` must win over `service.name` in `OTEL_RESOURCE_ATTRIBUTES`.** The SDK documents exactly this at `lib/src/environment/otel_env.dart:312` and `getServiceConfig` implements it correctly — the value is computed right, then overwritten downstream.
2. **Programmatic configuration must win over environment variables.** An explicit `serviceName:` argument silently losing to an env var is the more serious of the two.

## Why CI never caught it

CI runs with a clean environment, so `OTEL_RESOURCE_ATTRIBUTES` is unset and the clobber never happens. It only appears when a developer has the variable exported — which also explains why this looked like "tests assuming a clean environment".

Reproduce on `main`:

```bash
export OTEL_RESOURCE_ATTRIBUTES=service.name=dirty
dart test test/basic_test.dart
# Basic initialization works [E]
#   Expected: 'test-service'   <- passed explicitly to OTel.initialize
#     Actual: 'dirty'
```

## Affected tests

These fail under a dirty environment because of the bug, not because of bad assumptions:

- `test/basic_test.dart` — `Basic initialization works`
- `test/integration/environment_variables_test.dart` — `OTel.initialize uses environment variables when no params provided`, `explicit parameters override environment variables`
- `test/integration/resource_attributes_test.dart` — `EnvVarResourceDetector reads OTEL_RESOURCE_ATTRIBUTES from env`, `resource attributes available in OTel.initialize()`

A further set fails for unrelated ambient-environment reasons and needs separate triage: `unit/logs/logger_provider_test.dart`, `unit/logs/logs_config_test.dart`, `unit/metrics/export/otlp/http/otlp_http_metric_exporter_coverage_test.dart`, `unit/otel_default_endpoint_test.dart` (×2), `unit/otel_initialization_error_test.dart`, `unit/otel_sdk_test.dart`, `unit/sdk/resource_detector_test.dart`.

## Fix

Correct the resource precedence in `OTel.initialize` so the order is: explicit arguments > `OTEL_SERVICE_NAME` > `OTEL_RESOURCE_ATTRIBUTES` > defaults. Add spec-compliance tests covering each rung, and a CHANGELOG entry — this is user-visible behaviour in the reference implementation.

Not folded into #99, which stays a test-infrastructure-only change.

michaelbushe:
Fixes #103. Split out of #99, which stays test-infrastructure-only.

## The bug

`OTel.initialize` applies `OTEL_RESOURCE_ATTRIBUTES` **twice**:

1. through `EnvVarResourceDetector`, correctly ranked *below* the service resource (`otel.dart:293`); and
2. folded into `resourceAttributes` (`otel.dart:204-216`) and merged **last**, at the highest precedence (`otel.dart:307`).

The second copy overrode everything above it:

- **`OTEL_SERVICE_NAME` lost to `service.name` in `OTEL_RESOURCE_ATTRIBUTES`** — contradicting the spec, and contradicting `OTelEnv.getServiceConfig`, which already implements the ordering correctly and documents it at `otel_env.dart:312`.
- **An explicit `serviceName:` argument silently lost to an environment variable.** Programmatic configuration must outrank the environment; this is the more serious of the two.

## The fix

`getServiceConfig` already resolves `service.name`/`service.version` with the correct precedence, and `initialize` materialises both into the service resource — so the env-derived copy of *those two keys* is pure duplication. Excluding them from the attributes merged last restores the ladder. Every other key keeps its existing precedence, including non-service entries of the same variable.

Resulting order, highest first: **explicit argument → `OTEL_SERVICE_NAME` → `service.name` in `OTEL_RESOURCE_ATTRIBUTES` → default.**

| Case | Before | After |
|---|---|---|
| explicit `serviceName:` vs `OTEL_RESOURCE_ATTRIBUTES` | `from-ra` ❌ | `explicit-service` ✅ |
| `OTEL_SERVICE_NAME` vs `OTEL_RESOURCE_ATTRIBUTES` | `from-ra` ❌ | `from-env-svc` ✅ |
| `OTEL_RESOURCE_ATTRIBUTES` alone | `from-ra` ✅ | `from-ra` ✅ |
| clean environment | default ✅ | default ✅ |
| `deployment.environment` (non-service key) | preserved ✅ | preserved ✅ |

## Why CI never caught it

CI runs with a clean environment, so `OTEL_RESOURCE_ATTRIBUTES` is unset and the clobber never happens. It only surfaced for developers with the variable exported — which is exactly why it first looked like the tests were making bad assumptions.

## Tests

New `test/unit/otel_service_precedence_test.dart`: 12 tests covering each rung across **both** `detectPlatformResources` paths (true routes the variable through `EnvVarResourceDetector`, false does not). Confirmed they bite — with the fix reverted, **8 of the 12 fail**.

Also fixes three tests in `test/integration/resource_attributes_test.dart` that called `detect()` with no factory installed and threw `StateError` the moment their environment guard was satisfied. CI's clean environment meant those guards never fired, so the bodies had never actually run.

| Check | Result |
|---|---|
| Full suite, clean environment | 2016 passed, 13 skipped, 0 failed |
| `basic_test` + both integration files, dirty environment | 10 passed (was 5 failing) |
| New precedence tests with fix reverted | 8 of 12 fail, as intended |
| `dart analyze --fatal-infos` / `dart format` | clean |

Behaviour change is user-visible, so it carries a CHANGELOG entry under `1.1.0-beta.13-wip`.

### DIFF
--- a/lib/src/otel.dart
+++ b/lib/src/otel.dart
@@ -160,6 +160,10 @@ class OTel {
   }) async {
     // Has to run before anything logs OTLP headers.
     OTelEnv.applyHeaderLogAllowlist(otlpHeaderLogAllowlist);
+    // Apply OTEL_LOG_LEVEL before the env parsing below emits its debug
+    // output — otherwise an env-configured debug level misses exactly the
+    // lines that diagnose env configuration.
+    initializeLogging();
 
     // Apply environment variables only if parameters are not provided
     final envServiceName = serviceName == null
@@ -277,8 +281,6 @@ class OTel {
     _defaultTimeProvider = timeProvider;
     OTel.defaultTracerName = tracerName ?? _defaultTracerName;
     OTel.defaultTracerVersion = tracerVersion ?? defaultTracerVersion;
-    // Initialize logging from environment variables if needed
-    initializeLogging();
 
     final createdFactory = factoryFactory(
       apiEndpoint: endpoint ?? defaultEndpoint,
--- a/lib/src/otel.dart
+++ b/lib/src/otel.dart
@@ -200,8 +200,20 @@ class OTel {
     // Get otlpConfig for exporter creation later
     final otlpConfigForExporter = OTelEnv.getOtlpConfig(signal: 'traces');
 
-    // Get resource attributes from environment and merge with provided ones
-    final envResourceAttrs = OTelEnv.getResourceAttributes();
+    // Get resource attributes from environment and merge with provided ones.
+    //
+    // service.name and service.version are deliberately excluded here. They
+    // were already resolved into serviceName/serviceVersion above by
+    // getServiceConfig(), which applies the spec precedence -- OTEL_SERVICE_NAME
+    // outranks service.name in OTEL_RESOURCE_ATTRIBUTES -- and an explicit
+    // argument outranks both. These attributes are merged last, at the highest
+    // precedence, so leaving the service keys in would let
+    // OTEL_RESOURCE_ATTRIBUTES override the resolved values and silently beat
+    // programmatic configuration.
+    final envResourceAttrs =
+        Map<String, Object>.from(OTelEnv.getResourceAttributes())
+          ..remove(Service.serviceName.key)
+          ..remove(Service.serviceVersion.key);
     if (envResourceAttrs.isNotEmpty) {
       if (resourceAttributes != null) {
         // Merge with provided attributes - provided ones take precedence
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.1.0-beta.13-wip]
 
+### Fixed
+- **`OTEL_RESOURCE_ATTRIBUTES` no longer overrides `service.name` set by
+  higher-precedence sources** (#103). `OTel.initialize` applied that
+  variable twice: once through `EnvVarResourceDetector`, correctly ranked
+  below the service resource, and once folded into `resourceAttributes`
+  and merged last at the highest precedence. The second copy overrode
+  everything above it, so `OTEL_SERVICE_NAME` lost to `service.name` in
+  `OTEL_RESOURCE_ATTRIBUTES` (contradicting the spec, and the behaviour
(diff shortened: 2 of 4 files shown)


lib/src/otel.dart
```diff
@@ -216,8 +220,20 @@ class OTel {
...
+    // were already resolved into serviceName/serviceVersion above by
+    // getServiceConfig(), which applies the spec precedence -- OTEL_SERVICE_NAME
+    // outranks service.name in OTEL_RESOURCE_ATTRIBUTES -- and an explicit
+    // argument outranks both. These attributes are merged last, at the highest
+    // precedence, so leaving the service keys in would let
+    // OTEL_RESOURCE_ATTRIBUTES override the resolved values and silently beat
+    // programmatic configuration.
+    final envResourceAttrs =
+        Map<String, Object>.from(OTelEnv.getResourceAttributes())
+          ..remove(Service.serviceName.key)
```
michaelbushe, line 235: These removes smell very bad but we will rework this later.

PR #175, opened 2026-08-15
fix: Tracer.enabled returns false when no span processors are registered

### LINKED ISSUE #138: [BUG] Trace SDK: Tracer Enabled does not return false when no SpanProcessors are registered
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Trace SDK, Enabled
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/trace/sdk/#enabled
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/sdk.md#enabled

> `Enabled` MUST return `false` when either:
>
> - there are no registered [`SpanProcessors`](https://opentelemetry.io/docs/specs/otel/trace/sdk/#span-processor),
> - **Status**: [Development](https://opentelemetry.io/docs/specs/otel/document-status/) - `Tracer` is disabled ([`TracerConfig.enabled`](https://opentelemetry.io/docs/specs/otel/trace/sdk/#tracerconfig) is `false`).

## The problem

The `enabled` getter returns an instance field. That field starts as `true`. The getter never reads the processors of the provider.

A `TracerProvider` can hold an empty processor list. `tracer.enabled` still returns `true`.

Instrumentation uses `Enabled` as a fast check before expensive work. In this state the instrumentation pays the full cost of span creation for telemetry that reaches nothing.

## Offending code

`dartastic_opentelemetry/lib/src/trace/tracer.dart:37`

```dart
bool _enabled = true;
```

`dartastic_opentelemetry/lib/src/trace/tracer.dart:81-82`

```dart
@override
bool get enabled => _enabled;
```

The empty list is at `dartastic_opentelemetry/lib/src/trace/tracer_provider.dart:32`.

## How to correct it

Return `false` when the provider holds no processor:

```dart
bool get enabled => _enabled && _provider.spanProcessors.isNotEmpty;
```

NOTE: The second condition needs `TracerConfig`, which does not exist yet. That part has the status **Development**. This issue covers the first condition only.

## Notes

- Source finding files: `fable-ultra/trace-sdk-11-enabled-ignores-processors.md`, `fable-medium/trace-sdk-enabled-ignores-processors.md`
- Parent issue: #95

abidiahmedcom:
## Summary

Per the [OpenTelemetry Trace SDK spec](https://opentelemetry.io/docs/specs/otel/trace/sdk/#enabled), `Enabled` MUST return `false` when there are no registered SpanProcessors. This PR makes `Tracer.enabled` reflect that requirement.

### Problem

The `enabled` getter only checked the internal `_enabled` flag, which starts as `true`. A `TracerProvider` can hold an empty processor list (e.g. `OTEL_SDK_DISABLED` / bare named providers), yet `tracer.enabled` still returned `true`. Instrumentation uses `Enabled` as a fast check before expensive work, so in this state it paid the full cost of span creation for telemetry that reaches nothing.

### Fix

`Tracer.enabled` now returns `false` when the provider holds no span processors:

```dart
bool get enabled => _enabled && _provider.spanProcessors.isNotEmpty;
```

The explicit `enabled` setter is preserved and still controls the flag independently.

### Tests

Added a test verifying that a tracer from a provider with no registered span processors reports `enabled == false`, and flips to `true` once a processor is added.

- Fixes #138
- Validation: `dart test test/unit/trace/tracer_test.dart` (10 passed). Note: `test/unit/context/context_propagation_test.dart` failures are pre-existing on `main` (verified via `git stash`).

### DIFF
--- a/lib/src/trace/tracer.dart
+++ b/lib/src/trace/tracer.dart
@@ -79,8 +79,7 @@ class Tracer implements APITracer {
   set attributes(Attributes? attributes) => _delegate.attributes = attributes;
 
   @override
-  bool get enabled =>
-      _enabled && _provider.spanProcessors.isNotEmpty;
+  bool get enabled => _enabled && _provider.hasSpanProcessors;
 
   @override
   APISpan? get currentSpan => _delegate.currentSpan;
--- a/lib/src/trace/tracer.dart
+++ b/lib/src/trace/tracer.dart
@@ -79,7 +79,8 @@ class Tracer implements APITracer {
   set attributes(Attributes? attributes) => _delegate.attributes = attributes;
 
   @override
-  bool get enabled => _enabled;
+  bool get enabled =>
+      _enabled && _provider.spanProcessors.isNotEmpty;
 
   @override
   APISpan? get currentSpan => _delegate.currentSpan;
--- a/lib/src/trace/tracer_provider.dart
+++ b/lib/src/trace/tracer_provider.dart
@@ -219,7 +219,7 @@ class TracerProvider implements APITracerProvider {
   /// Whether any span processors are registered.
   ///
   /// Unlike [spanProcessors], this does not allocate a new unmodifiable list,
-  /// making it suitable for hot paths such as [SDKTracer.enabled].
+  /// making it suitable for hot paths such as [Tracer.enabled].
   bool get hasSpanProcessors => _spanProcessors.isNotEmpty;
 
   /// Ensures the resource for this provider is properly set.
--- a/lib/src/trace/tracer_provider.dart
+++ b/lib/src/trace/tracer_provider.dart
@@ -216,6 +216,12 @@ class TracerProvider implements APITracerProvider {
   /// @return An unmodifiable list of all span processors
   List<SpanProcessor> get spanProcessors => List.unmodifiable(_spanProcessors);
 
+  /// Whether any span processors are registered.
+  ///
+  /// Unlike [spanProcessors], this does not allocate a new unmodifiable list,
+  /// making it suitable for hot paths such as [SDKTracer.enabled].
+  bool get hasSpanProcessors => _spanProcessors.isNotEmpty;
+
   /// Ensures the resource for this provider is properly set.
   ///
   /// If no resource has been set, the default resource will be used.
--- a/test/unit/trace/tracer_provider_test.dart
+++ b/test/unit/trace/tracer_provider_test.dart
@@ -136,8 +136,6 @@ void main() {
     });
 
     test('hasSpanProcessors reflects registered processors', () {
-      expect(tracerProvider.hasSpanProcessors, isTrue);
-
       final bareProvider = OTel.addTracerProvider('bare-has-provider');
       expect(bareProvider.hasSpanProcessors, isFalse);
 
--- a/test/unit/trace/tracer_provider_test.dart
+++ b/test/unit/trace/tracer_provider_test.dart
@@ -135,6 +135,19 @@ void main() {
       expect(() => processors.add(MockSpanProcessor()), throwsUnsupportedError);
     });
 
+    test('hasSpanProcessors reflects registered processors', () {
+      expect(tracerProvider.hasSpanProcessors, isTrue);
+
+      final bareProvider = OTel.addTracerProvider('bare-has-provider');
+      expect(bareProvider.hasSpanProcessors, isFalse);
+
+      final mockProcessor3 = MockSpanProcessor();
+      bareProvider.addSpanProcessor(mockProcessor3);
+      expect(bareProvider.hasSpanProcessors, isTrue);
+
+      bareProvider.shutdown();
+    });
+
     test('ensureResourceIsSet sets resource if null', () {
       // Initially resource is default from OTel.initialize
       expect(tracerProvider.resource, isNotNull);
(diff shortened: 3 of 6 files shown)


yuzurihaaa:
Hi. Thank you for your contribution! I just created the list of issues here and still verifying stuff. I'll get back to you on this.

lib/src/trace/tracer.dart
```diff
@@ -79,7 +79,8 @@ class Tracer implements APITracer {
   set attributes(Attributes? attributes) => _delegate.attributes = attributes;
 
   @override
-  bool get enabled => _enabled;
+  bool get enabled =>
+      _enabled && _provider.spanProcessors.isNotEmpty;
```
yuzurihaaa: I realised that [spanProcessors](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/blob/f048243b14e1aec29b2a4a62e09ff6bc7948a80d/lib/src/trace/tracer_provider.dart#L217) here is creating a new `List.unmodifiable` when it is called. @harshitt13 @robert-northmind I'd like to get some opinion here if it is make sense for a getter in `TracerProvider` (we can call it `hasSpanProcessors`) to get `spanProcessors.isNotEmpty` value to avoid creating a new `List.unmodifiable` when `enabled` is called.

harshitt13: Yeah, I agree, since `enabled` is supposed to be a cheap fast-path check, allocating a new list every time kind of defeats the purpose. A simple `bool get hasSpanProcessors` on `TracerProvider` that just checks `_spanProcessors.isNotEmpty` directly makes sense to me.

yuzurihaaa: Hi @abidiahmedcom . Mind if I bother you with this change? In this [TracerProvider](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/blob/f048243b14e1aec29b2a4a62e09ff6bc7948a80d/lib/src/trace/tracer_provider.dart#L27) to add

```
bool hasSpanProcessors get => _spanProcessors.isNotEmpty;
```

then this function we can change it to

```
_enabled && _provider.hasSpanProcessors;
```


edit: on isEmpty -> isNotEmpty. It's already late night here.

abidiahmedcom:
Thanks @yuzurihaaa for the suggestion (and @harshitt13 for the confirmation) - implemented it.

Added `Bool get hasSpanProcessors => _spanProcessors.isNotEmpty;` to `TracerProvider` (no `List.unmodifiable` allocation), and updated `Tracer.enabled` to use it:

```dart
bool get enabled => _enabled && _provider.hasSpanProcessors;
```

Also added a test for `hasSpanProcessors` covering a bare provider (false) and one with a processor added (true). All 25 tracer/provider tests pass.

Note: the code block in your message reads `_spanProcessors.isEmpty` - I used `isNotEmpty` to match the getter's intent. Happy to flip the semantics if you'd prefer the name to signal empty instead

lib/src/trace/tracer_provider.dart
```diff
@@ -216,6 +216,12 @@ class TracerProvider implements APITracerProvider {
   /// @return An unmodifiable list of all span processors
   List<SpanProcessor> get spanProcessors => List.unmodifiable(_spanProcessors);
 
+  /// Whether any span processors are registered.
+  ///
+  /// Unlike [spanProcessors], this does not allocate a new unmodifiable list,
+  /// making it suitable for hot paths such as [SDKTracer.enabled].
```
yuzurihaaa:
```suggestion
  /// making it suitable for hot paths such as [Tracer.enabled].
```

test/unit/trace/tracer_provider_test.dart
```diff
@@ -135,6 +135,19 @@ void main() {
       expect(() => processors.add(MockSpanProcessor()), throwsUnsupportedError);
     });
 
+    test('hasSpanProcessors reflects registered processors', () {
+      expect(tracerProvider.hasSpanProcessors, isTrue);
```
yuzurihaaa: nit: remove this line, or move it to a separate test. setUp calls addSpanProcessor for this group, so tracerProvider already has processors. This line and the bareProvider lines test two different fixtures in one case.

The bareProvider part is the real test for hasSpanProcessors. It verifies the change from false to true.

yuzurihaaa:
[CI](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/actions/runs/31920330530/job/95117343045?pr=175) failing on verify formatting

lib/src/trace/tracer_provider.dart
abidiahmedcom: Thanks! Applied in 13df04a.

test/unit/trace/tracer_provider_test.dart
abidiahmedcom: Agreed - that line was redundant since setUp already adds processors for the group. Removed it in 13df04a; the bareProvider fixture now covers the false->true change.

abidiahmedcom:
Fixed - 	ool/backport_release.dart was unformatted on main. Ran dart format on it (commit 13df04a); the whole-repo format gate now passes locally. Note the analyze job also needs a fresh run since this PR now contains a new commit.

yuzurihaaa (approved):
LGTM

michaelbushe:
Thank you for the contribution @abidiahmedcom ! Thank you team (@yuzurihaaa, @harshitt13 ) for triage and reviews. Merging when green.

michaelbushe (approved):
👍🏻

Issue #228, opened 2026-08-15
[BUG] OTLP: exporters emit no identifying User-Agent header


yuzurihaaa:
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — OTLP Exporter, User agent
**Requirement level:** SHOULD
**URL:** https://opentelemetry.io/docs/specs/otel/protocol/exporter/#user-agent
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/protocol/exporter.md#user-agent

> OpenTelemetry protocol exporters SHOULD emit a User-Agent header to at a minimum identify the exporter, the language of its implementation, and the version of the exporter. For example, the Python OTLP exporter version 1.2.3 would report the following: `OTel-OTLP-Exporter-Python/1.2.3`

## The problem

The HTTP exporters build their headers from the user headers plus the content type and the content encoding. No `User-Agent` header is set.

A search for `User-Agent` and `userAgent` across the repository matches the web resource detector only.

Requests therefore go out with the generic default of the HTTP package, or with the browser user agent on the web.

The gRPC exporters use the grpc-dart default. A user cannot fix this through headers either, because grpc-dart strips `user-agent` as a reserved key.

## Offending code

`dartastic_opentelemetry/lib/src/trace/export/otlp/http/otlp_http_span_exporter.dart:126-139`

```dart
final headers = Map<String, String>.from(_config.headers);
...
headers['Content-Type'] = 'application/x-protobuf';
```

The metrics and logs HTTP exporters repeat this. The gRPC path is at `trace/export/otlp/otlp_grpc_span_exporter.dart:326-334`.

## How to correct it

1. Add a default header of the form `OTel-OTLP-Exporter-Dart/<package-version>` to every OTLP/HTTP request. A user header may add a distribution identifier in front.
2. Pass the same value through the channel options for the gRPC exporters.

## Notes

- Spec version: **v1.60.0**. The audit ran against v1.59.0. The quoted text above was confirmed word for word in the v1.60.0 source on 2026-08-15.
- Found by: Fable 5 (medium run)
- Source finding files: `fable-medium/otlp-user-agent-missing.md`
- Parent issue: #95
- Triage step 3a applies. A human must confirm this violation against the spec before work starts.

michaelbushe:
[...]

Issue #220, opened 2026-08-15
[BUG] OTLP: gRPC exporter defaults to port 4318 instead of 4317


yuzurihaaa:
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — OTLP Exporter, Configuration Options
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/protocol/exporter/#configuration-options
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/protocol/exporter.md#configuration-options

> **Endpoint (OTLP/gRPC)**: Target to which the exporter is going to send spans, metrics, or logs. ... Default: `http://localhost:4317`

## The problem

`OTel.initialize()` picks one endpoint before it looks at the protocol. That default is the OTLP/HTTP port.

With `OTEL_EXPORTER_OTLP_PROTOCOL=grpc` and no endpoint set, the gRPC exporter connects to port 4318. The spec default for gRPC is port 4317. Exports fail against a collector on the default ports.

The metrics path and the logs path receive the same already-resolved endpoint, so they inherit the fault.

The standalone gRPC config carries the correct default. The defect is in the wiring only.

## Offending code

`dartastic_opentelemetry/lib/src/otel.dart:77`, `:179`, and `:351-366`

```dart
static const defaultEndpoint = 'http://localhost:4318';
...
endpoint ??= defaultEndpoint;
...
if (protocol == 'grpc') {
  return OtlpGrpcSpanExporter(
    OtlpGrpcExporterConfig(
      endpoint: resolvedEndpoint,   // = http://localhost:4318 when no endpoint configured
```

## How to correct it

Pick the default after the code resolves the protocol. Use port 4317 for gRPC. Use port 4318 for the two HTTP protocols. Do this for each signal.

## Notes

- Spec version: **v1.60.0**. The audit ran against v1.59.0. The quoted text above was confirmed word for word in the v1.60.0 source on 2026-08-15.
- Found by: Fable 5 (both runs)
- Source finding files: `fable-medium/otlp-grpc-default-endpoint-port.md`, `fable-ultra/config-11-grpc-default-endpoint-port.md`
- Parent issue: #95
- Triage step 3a applies. A human must confirm this violation against the spec before work starts.

michaelbushe:
[...]

PR #238, opened 2026-08-16
Fix OTLP spec-compliance issues (#213, #220, #228)

### LINKED ISSUE #213: [BUG] Configuration: an empty environment variable value is not treated as unset
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — SDK environment variables, Parsing empty value
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#parsing-empty-value
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/configuration/sdk-environment-variables.md#parsing-empty-value

> The SDK MUST interpret an empty value of an environment variable the same way as when the variable is unset.

## The problem

`EnvironmentService.getValue` returns an empty string from the operating system environment as it is. The `--dart-define` path does filter an empty value. The `Platform.environment` fallback does not, and neither does the test override path.

The callers then treat an empty string as a value that the user set.

Two concrete failures follow.

With `OTEL_EXPORTER_OTLP_ENDPOINT` exported as an empty string, `OTel.initialize()` receives an empty endpoint and throws. The correct behavior is the default endpoint.

With `OTEL_SERVICE_NAME` exported as an empty string, the same method throws instead of using the default service name.

## Offending code

`dartastic_opentelemetry/lib/src/environment/environment_service_io.dart:72`

```dart
// Priority 2: Platform environment variables
value ??= io.Platform.environment[key];
```

The callers are at `environment/otel_env.dart:639`, `:87-104`, and `:317`. The throws are at `otel.dart:233` and `otel.dart:238`.

## How to correct it

Convert an empty string to null in `EnvironmentService.getValue`, or in `OTelEnv._getEnv`. Every consumer then reads empty as unset.

## Notes

- Source finding files: `fable-ultra/config-02-empty-env-value-not-unset.md`, `fable-medium/config-empty-env-treated-as-set.md`
- Parent issue: #95

abidiahmedcom:
Fixes #213, #220, #228 — three OTLP spec-compliance bugs (v1.60.0).

## Summary

**#213 — Empty env var values are now treated as unset.**
`EnvironmentService.getValue` (io + web) normalizes empty strings to `null`, so `OTEL_EXPORTER_OTLP_ENDPOINT=` no longer throws at initialize and falls back to the default endpoint; empty `OTEL_SERVICE_NAME`, protocol, log level, etc. all read as unset. (spec: configuration/sdk-environment-variables.md#parsing-empty-value)

**#220 — gRPC now defaults to port 4317, not 4318.**
The endpoint default is picked per signal AFTER the protocol is resolved: `http://localhost:4317` for gRPC, `http://localhost:4318` for the HTTP protocols. Adds `OTel.defaultGrpcEndpoint`. The trace path also re-resolves `secure` against the effective endpoint (matching metrics/logs and the spec's http-scheme precedence), so a default `http://localhost:4317` gRPC endpoint connects insecurely. (spec: protocol/exporter.md#configuration-options)

**#228 — OTLP requests now carry an identifying User-Agent.**
Every OTLP request sends `OTel-OTLP-Exporter-Dart/<package-version>` — as an HTTP header for the HTTP exporters and via `ChannelOptions.userAgent` for the gRPC exporters (grpc-dart strips user-supplied user-agent headers). A caller-supplied `user-agent` header is prepended to the default rather than replacing it, and is not consumed from the config between exports. (spec: protocol/exporter.md#user-agent)

## Tests
- New: `test/unit/otel_default_grpc_endpoint_test.dart`, `test/unit/otlp_user_agent_test.dart`, and empty-value cases in `otel_env_inprocess_test.dart`.
- Wire tests: default UA + prepended UA in the trace/metrics/logs HTTP suites; default UA on the gRPC metrics suite.
- `dart analyze` clean; `dart format` clean. Full suite: all pass except 8 pre-existing failures (4 `context_propagation_test.dart`, 4 `otlp_http_metric_exporter_coverage_test.dart`) — confirmed identical on pristine `f048243`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,14 +10,42 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+### Added
+
 - `TracerProvider.hasSpanProcessors` — allocation-free check for registered
   span processors.
+
+- **OTLP exporters send an identifying `User-Agent` header.** Per the OTLP
+  spec, OTLP requests SHOULD identify the exporter, language, and version.
+  Every OTLP request now carries `OTel-OTLP-Exporter-Dart/<version>` (HTTP
+  headers on the HTTP exporters; `ChannelOptions.userAgent` on the gRPC
+  exporters). A user-supplied `user-agent` header is prepended to the default
+  rather than replacing it (#228).
+
+- **`OTel.defaultGrpcEndpoint`** (`http://localhost:4317`), the OTLP/gRPC
+  default endpoint. The endpoint default is now picked per signal after the
+  protocol is resolved instead of defaulting everything to the HTTP port
+  4318 (#220).
 
 ### Fixed
 
 - `Tracer.enabled` now returns `false` when `TracerProvider` has no span
   processor(s) registered, per the Trace SDK spec, sparing span-creation cost
   when nothing is listening. Thanks to @abidiahmedcom (#138, #175).
+
+- **OTLP/gRPC exporters now default to port 4317, not 4318.** The OTLP spec
+  defaults the endpoint to `http://localhost:4317` for OTLP/gRPC and
+  `http://localhost:4318` for the two HTTP protocols. Previously a single
+  4318 default was applied before the protocol was known, so gRPC-only
+  deployments silently exported to the wrong port (#220).
+
+- **An empty environment variable value is treated as unset.** Per the spec,
+  an empty value of an environment variable MUST be interpreted the same way
+  as when the variable is unset. `EnvironmentService.getValue` now normalizes
+  empty strings to `null`, so every consumer (endpoint, protocol, service
+  name, log level, …) reads an empty value as unset (#213).
+
+Thanks to @abidiahmedcom; reported by @yuzurihaaa (#213, #220, #228).
 
 ## [1.1.0-beta.13] - 2026-08-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-### Added
-
 - `TracerProvider.hasSpanProcessors` — allocation-free check for registered
   span processors.
 
--- a/lib/src/export/otlp_user_agent.dart
+++ b/lib/src/export/otlp_user_agent.dart
@@ -1,0 +1,19 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+/// Default User-Agent the OTLP exporters send on every request.
+///
+/// Per the OTLP exporter spec (specification/protocol/exporter.md, "User
+/// agent"), OTLP exporters SHOULD emit a `User-Agent` header that at minimum
+/// identifies the exporter, the language of its implementation, and the
+/// version of the exporter. The HTTP exporters send this header directly; the
+/// gRPC exporters pass it through `ChannelOptions.userAgent`, because gRPC
+/// treats `user-agent` as a reserved header and would strip it from a
+/// user-supplied map.
+library;
+
+/// Hand-synced with the package version in pubspec.yaml.
+const String packageVersion = '1.1.0-beta.14-wip';
+
+/// The default User-Agent value, e.g. `OTel-OTLP-Exporter-Dart/1.1.0-beta.14-wip`.
+const String otlpUserAgent = 'OTel-OTLP-Exporter-Dart/$packageVersion';
--- a/lib/src/export/otlp_user_agent.dart
+++ b/lib/src/export/otlp_user_agent.dart
@@ -12,8 +12,8 @@
 /// user-supplied map.
 library;
 
-/// Hand-synced with the package version in pubspec.yaml.
-const String packageVersion = '1.1.0-beta.14-wip';
+import '../version.dart';
 
-/// The default User-Agent value, e.g. `OTel-OTLP-Exporter-Dart/1.1.0-beta.14-wip`.
+/// The default User-Agent value, e.g. `OTel-OTLP-Exporter-Dart/1.1.0-beta.14`.
+/// The version is [packageVersion], stamped by the release tooling (#249).
 const String otlpUserAgent = 'OTel-OTLP-Exporter-Dart/$packageVersion';
--- a/test/unit/environment/otel_env_inprocess_test.dart
+++ b/test/unit/environment/otel_env_inprocess_test.dart
@@ -328,4 +328,47 @@ void main() {
       expect(OTelEnv.getLogRecordLimits(), isEmpty);
     });
   });
+
+  group('parsing empty values (issue #213)', () {
+    test('an empty value reads as unset from getValue', () {
+      env({'OTEL_SERVICE_NAME': ''});
+      expect(EnvironmentService.instance.getValue('OTEL_SERVICE_NAME'), isNull);
+    });
+
+    test('an empty endpoint reads as unset', () {
+      env({'OTEL_EXPORTER_OTLP_ENDPOINT': ''});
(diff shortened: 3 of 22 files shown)


lib/src/export/otlp_user_agent.dart
```diff
@@ -0,0 +1,19 @@
...
+/// agent"), OTLP exporters SHOULD emit a `User-Agent` header that at minimum
+/// identifies the exporter, the language of its implementation, and the
+/// version of the exporter. The HTTP exporters send this header directly; the
+/// gRPC exporters pass it through `ChannelOptions.userAgent`, because gRPC
+/// treats `user-agent` as a reserved header and would strip it from a
+/// user-supplied map.
+library;
+
+/// Hand-synced with the package version in pubspec.yaml.
+const String packageVersion = '1.1.0-beta.14-wip';
```
michaelbushe: This hand-synced const will drift: `tool/release.dart` rewrites only pubspec/CHANGELOG/README, so the released `1.1.0-beta.14` would advertise `OTel-OTLP-Exporter-Dart/1.1.0-beta.14-wip`, and the `0.9.x` backport would advertise the beta's version. #250 adds a release-tooling-stamped `lib/src/version.dart` for exactly this (context in #249). Once it lands on main, please rebase and derive from the shared const:

```dart
import '../version.dart';

const String otlpUserAgent = 'OTel-OTLP-Exporter-Dart/$packageVersion';
```

dropping the local `packageVersion` here.

test/unit/otlp_user_agent_test.dart
```diff
@@ -0,0 +1,26 @@
...
+void main() {
+  group('otlpUserAgent (issue #228)', () {
+    test('matches the spec format OTel-OTLP-Exporter-Dart/<package-version>',
+        () {
+      expect(otlpUserAgent, equals('OTel-OTLP-Exporter-Dart/$packageVersion'));
+      expect(otlpUserAgent, startsWith('OTel-OTLP-Exporter-Dart/'));
+      expect(otlpUserAgent, isNot(endsWith('/')));
+    });
+
+    test('package version matches pubspec.yaml', () {
```
michaelbushe: This test never reads pubspec.yaml — it compares the const to another hardcoded literal, so it can't catch the drift it's named for; it would pass forever regardless of what pubspec says. After rebasing onto #250, `test/unit/version_sync_test.dart` does the real pubspec comparison (read from disk), and this test can shrink to the format assertions above, or be dropped.

test/unit/otel_default_grpc_endpoint_test.dart
```diff
@@ -0,0 +1,111 @@
...
+
+  test('grpc protocol defaults to port 4317 for traces', () async {
+    EnvironmentService.testOverrides = {'OTEL_EXPORTER_OTLP_PROTOCOL': 'grpc'};
+    await OTel.initialize(serviceName: 'grpc-traces-test');
+
+    final spanProcessor =
+        OTel.tracerProvider().spanProcessors.first as BatchSpanProcessor;
+    expect(spanProcessor.exporter, isA<OtlpGrpcSpanExporter>());
+
+    expect(logs.join('\n'), contains('endpoint: http://localhost:4317'));
```
michaelbushe, line 58: Non-blocking: these assertions verify what the code *logs*, not where the bytes go — they'd stay green if the log message and the actual dial target ever diverged, and they break on any log-format change. The per-signal breadth here is valuable and worth keeping; `test/unit/otlp_spec_compliance_behavior_test.dart` (added in cc23154) complements it with a live in-process gRPC server bound to 4317 asserting the export actually arrives.

test/unit/environment/otel_env_inprocess_test.dart
```diff
@@ -328,4 +328,47 @@ void main() {
       expect(OTelEnv.getLogRecordLimits(), isEmpty);
     });
   });
+
+  group('parsing empty values (issue #213)', () {
```
michaelbushe, line 332: Good per-key coverage of the env layer. One gap: nothing exercises the symptom reported in #213 — `OTel.initialize()` throwing on an empty `OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_SERVICE_NAME`. If an empty value ever reaches `initialize` through a path that bypasses `getValue` normalization, this group stays green while the issue regresses. The initialize-level tests in `test/unit/otlp_spec_compliance_behavior_test.dart` (cc23154) now cover that end of the chain.

CHANGELOG.md
```diff
@@ -8,6 +8,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.1.0-beta.14-wip]
 
+### Added
+
+- **OTLP exporters send an identifying `User-Agent` header.** Per the OTLP
```
michaelbushe, line 16: Two things for the CHANGELOG: (1) please add credit per the house style (see the beta.13 entries) — the fixes are yours and the issues were reported by @yuzurihaaa, e.g. `Thanks to @abidiahmedcom; reported by @yuzurihaaa (#213, #220, #228).` (2) Heads-up: main's `[1.1.0-beta.14-wip]` section gained the #175 entry after this branch was cut, so expect a small conflict here on rebase.

michaelbushe (changes requested):
Thanks @abidiahmedcom — all three fixes are correct and were verified independently: I wrote black-box tests from the issues and the spec *before* reading this diff, confirmed each red on pre-fix main (initialize throwing, exports dialing 4318, `dart-grpc/2.0.0` and `Dart/3.13 (dart:io)` user-agents on the wire), and all pass on this branch. Those tests are now on the branch in cc23154 (`test/unit/otlp_spec_compliance_behavior_test.dart` plus a logs-gRPC User-Agent test), alongside your per-signal suites — the two styles complement each other.

Requesting changes for one structural item plus housekeeping:

1. **Version sync** — the hand-synced `packageVersion` in `otlp_user_agent.dart` will ship stale (`-wip` in releases, beta version in 0.9.x backports), and its guard test compares literal-to-literal so it can't catch drift. #250 (on main shortly, context in #249) adds a release-tooling-stamped `lib/src/version.dart`; please rebase onto main and derive `otlpUserAgent` from that shared const. Inline comments have the details.
2. **CHANGELOG** — add credit lines (you + @yuzurihaaa who reported the issues), and expect a small merge conflict in the `-wip` section from the #175 entry when you rebase.

The log-string-assertion note on `otel_default_grpc_endpoint_test.dart` is non-blocking.

abidiahmedcom:
Thanks for the thorough review, @michaelbushe. Here's what I've addressed in the latest push (rebased onto `25f7e81`):

**Done:**

- **CHANGELOG** — Added credit lines per house style: `Thanks to @abidiahmedcom; reported by @yuzurihaaa (#213, #220, #228)`. Resolved the merge conflict from #175's entry in the `-wip` section.

- **Version test** — Removed the redundant `package version matches pubspec.yaml` test that compared a const to a hardcoded literal (as you noted, it could never catch drift). Replaced with a simple non-empty format assertion. The real pubspec comparison will come from `version_sync_test.dart` once #250 lands.

- **Rebased onto main** — Branch now includes #175 (`25f7e81`). No other conflicts.

**Pending on #250:**

The version sync (`otlpUserAgent` deriving from `lib/src/version.dart`) is blocked on #250 landing. Once it merges, I'll do a final rebase and wire up the shared const, dropping the local `packageVersion` from `otlp_user_agent.dart`.

**Verification:** `dart analyze --fatal-infos` clean, `dart format` clean (354 files, 0 changed), all touched test suites pass.

michaelbushe:
@abidiahmedcom You were fast, I was slow.  Thanks for making the adjustment. I merged #250, rebased and fixed up what you were forced to dangle.  Consider joining the Dart and Flutter OTel SIG! 
[Call For Contributors](https://opentelemetry.io/blog/2026/dart-flutter-opentelemetry/)
[Discussion](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/discussions/239)

PR #240, opened 2026-08-18
fix: preserve existing baggage when header is unparseable

### LINKED ISSUE #200: [BUG] Propagation: an unparseable baggage header overwrites existing Baggage
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Propagators API, Extract
**URL:** https://opentelemetry.io/docs/specs/otel/context/api-propagators/#extract
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/context/api-propagators.md#extract

> If a value can not be parsed from the carrier, for a cross-cutting concern, the implementation MUST NOT throw an exception and MUST NOT store a new value in the `Context`, in order to preserve any previously existing valid value.

## The problem

`extract` builds an entry map, then always stores a `Baggage` object into the Context.

When the header is present but no entry survives parsing, that object is empty. The guards can skip every pair. Three headers that produce this today are `"garbage"`, `"=x"`, and `"a=b=c"`.

`Context.withBaggage` replaces the baggage slot completely. An incoming Context that already carried valid baggage therefore loses it. One malformed header wipes it out.

This is the exact case the spec forbids. The code stores a new value on a parse failure, and it fails to keep the existing valid value.

## Offending code

`dartastic_opentelemetry/lib/src/context/propagation/w3c_baggage_propagator.dart:47-70`

```dart
final entries = <String, BaggageEntry>{};
final pairs = value.split(',');
for (final pair in pairs) {
  ...
  if (keyValue.length != 2) continue;   // all entries can be skipped
  ...
}

final baggage = OTel.baggage(entries);  // possibly empty
return context.withBaggage(baggage);    // unconditionally stored
```

The replacement happens at `dartastic_opentelemetry_api/lib/src/api/context/context.dart:160`.

## How to correct it

Return the original Context when the entry map is empty:

```dart
if (entries.isEmpty) return context;
return context.withBaggage(OTel.baggage(entries));
```

NOTE: The `a=b=c` case is a parse defect of its own. See the `=` in value issue in this same section. Fixing that one reduces how often this bug fires, but does not remove it.

## Notes

- Source finding files: `fable-ultra/context-09-baggage-extract-overwrites-on-parse-failure.md`, `fable-medium/baggage-extract-overwrites-on-parse-failure.md`, `grok/context/002-baggage-extract-overwrites.md`
- Parent issue: #95

vjymisal0:
Fixes #200. Ensures that when the baggage header is unparseable and produces no valid entries, we return the existing context unmodified so that we don't accidentally drop valid baggage.

### DIFF
--- a/lib/src/context/propagation/w3c_baggage_propagator.dart
+++ b/lib/src/context/propagation/w3c_baggage_propagator.dart
@@ -35,7 +35,9 @@ class W3CBaggagePropagator
     TextMapGetter<String> getter,
   ) {
     final value = getter.get(_baggageHeader);
-    OTelLog.debug('Extracting baggage: $value');
+    if (OTelLog.isDebug()) {
+      OTelLog.debug('Extracting baggage: $value');
+    }
     if (value == null || value.isEmpty) {
       // Propagators API spec: extract returns the passed context, updated
       // with extracted values — and unchanged when there is nothing to
--- a/lib/src/context/propagation/w3c_baggage_propagator.dart
+++ b/lib/src/context/propagation/w3c_baggage_propagator.dart
@@ -66,6 +66,8 @@ class W3CBaggagePropagator
       entries[key] = OTel.baggageEntry(value, metadata);
     }
 
+    if (entries.isEmpty) return context;
+
     final baggage = OTel.baggage(entries);
     return context.withBaggage(baggage);
   }
--- a/test/unit/baggage/w3c_baggage_propagator_test.dart
+++ b/test/unit/baggage/w3c_baggage_propagator_test.dart
@@ -127,10 +127,9 @@ void main() {
       final getter = TestTextMapGetter(carrier);
 
       final extractedContext = propagator.extract(context, carrier, getter);
-      final extractedBaggage = extractedContext.baggage;
 
-      expect(extractedBaggage, isA<Baggage>());
-      expect(extractedBaggage!.getAllEntries(), isEmpty);
+      expect(extractedContext, same(context));
+      expect(extractedContext.baggage, isNull);
     });
 
     test('unparseable baggage header preserves existing baggage', () {
--- a/test/unit/baggage/w3c_baggage_propagator_test.dart
+++ b/test/unit/baggage/w3c_baggage_propagator_test.dart
@@ -133,6 +133,21 @@ void main() {
       expect(extractedBaggage!.getAllEntries(), isEmpty);
     });
 
+    test('unparseable baggage header preserves existing baggage', () {
+      final existingBaggage = OTel.baggage({'test': OTel.baggageEntry('value')});
+      final contextWithExisting = context.withBaggage(existingBaggage);
+
+      final carrier = {'baggage': 'invalid_format'};
+      final getter = TestTextMapGetter(carrier);
+
+      final extractedContext = propagator.extract(contextWithExisting, carrier, getter);
+      final extractedBaggage = extractedContext.baggage;
+
+      expect(extractedContext, same(contextWithExisting));
+      expect(extractedBaggage, isNotNull);
+      expect(extractedBaggage!.getEntry('test')?.value, equals('value'));
+    });
+
     test(
         'extract without a baggage header returns the passed context'
         ' unchanged', () {
--- a/test/unit/baggage/w3c_baggage_propagator_test.dart
+++ b/test/unit/baggage/w3c_baggage_propagator_test.dart
@@ -133,13 +133,15 @@ void main() {
     });
 
     test('unparseable baggage header preserves existing baggage', () {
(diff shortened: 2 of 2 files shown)


test/unit/baggage/w3c_baggage_propagator_test.dart
```diff
@@ -133,6 +133,21 @@ void main() {
       expect(extractedBaggage!.getAllEntries(), isEmpty);
```
harshitt13: This test conflicts with the fix below; `context` here has no baggage attached, so with the new `if (entries.isEmpty) return context;` in the propagator, `extractedContext.baggage` will now be `null` instead of a non-null empty `Baggage`. This is why `test (stable)` is failing in CI.

Can you update this test to match the new behavior?

```dart
test('handles invalid baggage format', () {
  final carrier = {'baggage': 'invalid format'};
  final getter = TestTextMapGetter(carrier);
  final extractedContext = propagator.extract(context, carrier, getter);
  expect(extractedContext, same(context));
  expect(extractedContext.baggage, isNull);
});
```

harshitt13 (changes requested):

lib/src/context/propagation/w3c_baggage_propagator.dart
michaelbushe: This was not in your scope but please surround this with the missing `if (OTelLog.isDebug()) {` so that propogators don't evaluate $value and make a string unnecessarily.  Thank you.

michaelbushe:
@vjymisal0 Thank you for your contribution.  Please be sure to read CONTRIBUTORS.md
Run `dart analyze` and `dart format` and `./tool/coverage.sh` and ensure tests pass and coverage increases or stays the same, thank you.

vjymisal0:
Good catch, thanks. Updated the invalid-format test to expect the new behavior (context returned unchanged, baggage null) instead of the old empty-Baggage expectation. Ran the full file locally and all 8 tests pass now.

harshitt13:
@vjymisal0 Please run `dart format .`

harshitt13:
Thanks for the work @vjymisal0!

However, I think there are a couple of things we should address to fully close out #200:

### 1. `split('=')` drops valid entries where the value contains `=`

The W3C Baggage spec ([§3.3.1.3](https://www.w3.org/TR/baggage/#value)) explicitly states:

> `value` MAY contain any number of the equal sign (`U+003D`) code points. Parsers **MUST NOT** assume that the equal sign is only used to separate `key` and `value`.

The current parsing on [line 53–54](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/blob/main/lib/src/context/propagation/w3c_baggage_propagator.dart#L53-L54):

```dart
final keyValue = trimmedPair.split('=');
if (keyValue.length != 2) continue;
```

…splits on *every* `=`, so a header like `a=b=c` produces `['a', 'b', 'c']` (length 3) and gets silently dropped. Per the spec it should parse as key=`a`, value=`b=c`.

Issue #200 calls this out directly:

> NOTE: The `a=b=c` case is a parse defect of its own.

A fix would be to split only on the first `=`:

```dart
final eqIndex = trimmedPair.indexOf('=');
if (eqIndex <= 0) continue; // no '=' or empty key
final key = _decodeComponent(trimmedPair.substring(0, eqIndex).trim());
if (key.isEmpty) continue;

final rawValue = trimmedPair.substring(eqIndex + 1);
final valueAndMetadata = rawValue.split(';');
```

### 2. Test coverage for the specific cases in #200

The issue lists three headers that trigger the bug: `"garbage"`, `"=x"`, and `"a=b=c"`. Currently only the `"garbage"` case is covered (via `'invalid_format'`). Could you add tests for:

- **`=x`** — empty key, should skip and preserve existing baggage
- **`a=b=c`** — once the parsing fix above is in, should parse as `key=a, value=b=c`
- **Mixed valid + invalid** — e.g. `valid=entry,garbage` — valid entries should still be extracted
- **All-invalid with existing baggage** — e.g. `garbage,=x` on a context that already has baggage — should preserve the existing baggage

michaelbushe (changes requested):

michaelbushe:
@vjymisal0  #261 closed #200.  My bad, I didn't realize this was the same.  Please check your work against the latest and see if there's anythign else you'd want to bring forward.  Thanks again.

test/unit/baggage/w3c_baggage_propagator_test.dart
vjymisal0: this was already fixed in 61982bc, the assertion now checks extractedContext.baggage is null - test passes

lib/src/context/propagation/w3c_baggage_propagator.dart
vjymisal0: wrapped it in isDebug() in 7bc2af2

PR #251, opened 2026-08-21
feat: config records for OTelEnv & strict boolean parsing

### LINKED ISSUE #97: [FEATURE] Dart Records for env vars
As described in https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/pull/66/changes#r3741235964

otel_env.dart should return Dart Records for every getXXXConfig function.

**Solution**
For example:
Instead of:
`static Map<String, dynamic> getBlrpConfig() {`

```
typedef BlrpEnvironmentalVariables = ({
  Duration? scheduleDelay,
  Duration? exportTimeout,
  int? maxQueueSize,
  int? maxExportBatchSize,
});

`static BlrpEnvironmentalVariables getBlrpConfig() {`

```

### LINKED ISSUE #212: [BUG] Configuration: boolean env vars accept 1, yes, and on as true
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — SDK environment variables, Boolean
**URL:** https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#boolean
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/configuration/sdk-environment-variables.md#boolean

> Any value that represents a Boolean MUST be set to true only by the case-insensitive string `"true"`, meaning `"True"` or `"TRUE"` are also accepted, as true. An implementation MUST NOT extend this definition and define additional values that are interpreted as true. Any value not explicitly defined here as a true value, including unset and empty values, MUST be interpreted as false.

## The problem

The boolean parsers treat `1`, `yes`, and `on` as true. The spec allows the case-insensitive string `true` only.

These parsers drive real switches. `OTEL_SDK_DISABLED` uses one. `OTEL_EXPORTER_OTLP_INSECURE` and the per-signal variants use the other.

`OTEL_SDK_DISABLED=1` therefore disables the SDK. `OTEL_EXPORTER_OTLP_INSECURE=yes` therefore turns off TLS. The spec forbids both readings.

No warning appears when a value falls back either.

## Offending code

`dartastic_opentelemetry/lib/src/environment/otel_env.dart:650`

```dart
static bool _getEnvBool(String name) {
  final value = _getEnv(name)?.toLowerCase();
  return value == '1' || value == 'true' || value == 'yes' || value == 'on';
}
```

`_getEnvBoolNullable` at `otel_env.dart:663` accepts the same extra values, and treats `0`, `no`, and `off` as false. The call sites are at `otel_env.dart:371` and `otel_env.dart:164-174`.

## How to correct it

1. Treat only the case-insensitive string `true` as true.
2. Treat every other value as false, including `1`, `yes`, and `on`.
3. Write a warning when a value is not `true`, not `false`, not empty, and not unset.

CAUTION: `OTEL_EXPORTER_OTLP_INSECURE` controls TLS. A deployment that sets `yes` today turns TLS off and will start using TLS after this fix. Note that in the changelog.

## Notes

- Source finding files: `fable-ultra/config-01-boolean-env-extra-true-values.md`, `fable-medium/config-boolean-parsing.md`, `grok/environment/001-boolean-truthy-extension.md`
- Parent issue: #95

harshitt13:
Fixes #97.
Fixes #212.

Refactors configuration parsing in `OTelEnv` to return strongly-typed Dart Records instead of untyped `Map<String, dynamic>`. This provides compile-time safety and eliminates unsafe dynamic casts across exporter configurations. 

Additionally, this PR brings the SDK into compliance with the OpenTelemetry v1.60.0 specification for boolean environment variables, strictly accepting only `"true"` and `"false"`.

### Changes
* **Added Record Typedefs:** Created `OtlpEnvironmentValues`, `BspEnvironmentValues`, `BlrpEnvironmentValues`, `ServiceEnvironmentValues`, and `LogRecordLimitsEnvironmentValues`.
* **Refactored `OTelEnv`:** Updated all `getXXXConfig` methods to return these new records.
* **Updated Consumers:** Replaced map lookups (`config['key'] as Type`) with record property access (`config.key`) across exporter configurations (`otel.dart`, `logs_config.dart`, `metric_config.dart`).
* **Strict Boolean Parsing:** Added `_getEnvBool` and `_getEnvBoolNullable` in `OTelEnv` to enforce the spec requirement that booleans must be strictly `"true"` or `"false"` (case-insensitive). Invalid values now gracefully default/warn rather than implicitly enabling features.
* **Test Infrastructure:** Resolved `jsonDecode` output parsing errors in `helpers/` to properly isolate JSON outputs from `stdout` warnings.

### Breaking Change
* `OTelEnv` is exported from the public barrel. Changing every `getXXXConfig` return type from `Map<String, dynamic>` to a record is a breaking API change (documented in the `CHANGELOG.md`).
  * **Migration hint:** Update map accesses to record property accesses (e.g., `config['endpoint'] as String?` → `config.endpoint`).

### DIFF
--- a/lib/src/otel.dart
+++ b/lib/src/otel.dart
@@ -162,22 +162,20 @@ class OTel {
     OTelEnv.applyHeaderLogAllowlist(otlpHeaderLogAllowlist);
 
     // Apply environment variables only if parameters are not provided
-    final envServiceName = serviceName == null
-        ? OTelEnv.getServiceConfig()['serviceName'] as String?
-        : null;
-    final envServiceVersion = serviceVersion == null
-        ? OTelEnv.getServiceConfig()['serviceVersion'] as String?
-        : null;
+    final envServiceConfig = OTelEnv.getServiceConfig();
+    final envServiceName =
+        serviceName == null ? envServiceConfig.serviceName : null;
+    final envServiceVersion =
+        serviceVersion == null ? envServiceConfig.serviceVersion : null;
 
     serviceName ??= envServiceName;
     serviceVersion ??= envServiceVersion;
 
     final otlpConfig = (endpoint == null || secure == null)
         ? OTelEnv.getOtlpConfig(signal: 'traces')
-        : <String, dynamic>{};
-    final envEndpoint =
-        endpoint == null ? otlpConfig['endpoint'] as String? : null;
-    final envInsecure = secure == null ? otlpConfig['insecure'] as bool? : null;
+        : null;
+    final envEndpoint = endpoint == null ? otlpConfig?.endpoint : null;
+    final envInsecure = secure == null ? otlpConfig?.insecure : null;
 
     endpoint ??= envEndpoint;
     secure = OTelEnv.resolveOtlpSecure(
@@ -286,8 +284,7 @@ class OTel {
     _installGlobalPropagator();
 
     if (OTelLog.isDebug()) {
-      final traceProtocol =
-          otlpConfigForExporter['protocol'] as String? ?? 'http/protobuf';
+      final traceProtocol = otlpConfigForExporter.protocol ?? 'http/protobuf';
       OTelLog.debug(
         'OTel initialized with endpoint: '
         '${endpoint ?? (traceProtocol == 'grpc' ? defaultGrpcEndpoint : defaultEndpoint)}, '
@@ -359,8 +356,7 @@ class OTel {
         // spanProcessor remains null and no processor is added.
       } else {
         // Determine protocol - default to http/protobuf if not set
-        final protocol =
-            otlpConfigForExporter['protocol'] as String? ?? 'http/protobuf';
+        final protocol = otlpConfigForExporter.protocol ?? 'http/protobuf';
 
         // Capture the resolved values: promotion of the nullable
         // parameters does not carry into the closure. The endpoint default is
@@ -372,7 +368,7 @@ class OTel {
         final resolvedEndpoint = endpoint ??
             (protocol == 'grpc' ? defaultGrpcEndpoint : defaultEndpoint);
         final resolvedSecure = OTelEnv.resolveOtlpSecure(
-          envInsecure: otlpConfigForExporter['insecure'] as bool?,
+          envInsecure: otlpConfigForExporter.insecure,
           endpoint: resolvedEndpoint,
           fallback: secure,
         );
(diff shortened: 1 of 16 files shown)


lib/src/otel.dart
```diff
@@ -153,22 +153,30 @@ class OTel {
...
+            endpoint: null,
+            protocol: null,
+            headers: null,
+            insecure: null,
+            timeout: null,
+            compression: null,
+            certificate: null,
+            clientKey: null,
+            clientCertificate: null,
+          ) as OtlpEnvironmentValues;
```
michaelbushe: Style, non-blocking: this 11-line all-null record literal grows every time a field is added, and the `as` cast is unnecessary (`({Null endpoint, …})` is already a subtype of `OtlpEnvironmentValues`). A nullable record reads better and needs no placeholder:

```dart
final otlpConfig = (endpoint == null || secure == null)
    ? OTelEnv.getOtlpConfig(signal: 'traces')
    : null;
final envEndpoint = endpoint == null ? otlpConfig?.endpoint : null;
final envInsecure = secure == null ? otlpConfig?.insecure : null;
```

michaelbushe (changes requested):
Thanks @harshitt13 — this is a solid refactor and we want it in **ahead of #72 and #74**, so they can adopt the record pattern for their new config surface instead of adding more `Map<String, dynamic>`.

Verified locally: audited the conversion line-by-line (every `config['x'] as T? ?? d` → `config.x ?? d` is faithful, incl. the BSP `is Duration` → `!= null` checks), `dart analyze` clean, and the environment/BSP/logs-config suites pass (+312). Nice touch extending the existing `BlrpEnvironmentValues` precedent rather than inventing a new shape.

Two asks before merge:

1. **Resolve conflicts onto current main.** Only two files conflict — `lib/src/logs/export/logs_config.dart` and `lib/src/metrics/export/metric_config.dart` — both from #238 landing after your branch point. One care point: #238 moved protocol resolution **above** the endpoint default in both files (the gRPC-4317 fix, #220), so when resolving, keep main's structure and apply only the map→record access conversion on top. `lib/src/otel.dart` and the tests auto-merge cleanly.

2. **CHANGELOG entry.** `OTelEnv` is exported from the public barrel, so changing every `getXXXConfig` return type from `Map<String, dynamic>` to records is a **breaking API change**. Please add a `### Changed` entry under `## [1.1.0-beta.14-wip]` saying so, with a one-line migration hint (`config['endpoint'] as String?` → `config.endpoint`).

One inline style suggestion on the null-record placeholder, non-blocking.

Heads-up for after the rebase: main also gained `test/unit/otlp_spec_compliance_behavior_test.dart` and the user-agent/version-sync tests — they should all stay green through this refactor and are a good no-behavior-drift check for it.

michaelbushe:
Nice work @harshitt13.  This code looks so much better than before.  Thanks for the idea, @robert-northmind.

michaelbushe (approved):

PR #259, opened 2026-08-22
fix: populate browser.* resource attributes on web

### LINKED ISSUE #190: [BUG] WebResourceDetector: all browser.* resource attributes are silently empty on web (invalid @JS bindings)
On web, WebResourceDetector.detect() throws on its first statement involving _isMobile(), the catch swallows it, and every browser resource attribute falls back to ''. Affects both DDC and dart2js — debug and release.

Lost attributes: browser.language, browser.platform, browser.vendor, browser.mobile, browser.languages, and  user_agent.original.

Console output at init:
  
[ERROR] Error detecting web resources: TypeError: Cannot read properties of undefined (reading 'test(window')

Root cause

In lib/src/resource/web_detector_impl.dart, two externals declare a JS function body as the @JS argument:

@JS('function() { '
      'return /Mobile|Android|iPhone|iPad|iPod|Windows Phone/i.test(window.navigator.userAgent) ? "true" : "false";'
      '}')
external String _isMobile();

@JS('...') names a global binding path, not a function body. The compiler splits the string on . and walks the parts as property lookups. dart2js emits it literally:

["function() { return /Mobile|Android|iPhone|iPad|iPod|Windows Phone/i"]["test(window"].navigator['userAgent) ? "true" : …

Hence reading 'test(window' — it is partway through walking the source text as a property chain. _getLanguagesString() is declared the same way and would fail identically; _isMobile() simply throws first.

@JS('window.navigator') on _navigator is correct usage and is not part of the problem.

Reproduce

Run the dart-otel-reference-demo and launch the flutter web app according to the instructions.  

Or run any Flutter web app that calls OTel.initialize() and observe the error at startup; the emitted resource carries empty  browser.* values.

Suggested fix
  
Neither helper needs JS interop — delete both externals and compute in Dart from the existing NavigatorJS extension:
  - browser.mobile — a RegExp over nav.userAgent.
  - browser.languages — add languages to NavigatorJSExtension as JSArray<JSString>? and join in Dart.

The blanket catch made this silent: a detector failure currently degrades to empty strings  rather than omitting the attributes, so the resource looks populated-but-blank instead of absent. Since this is in initialize(), it could throw (but maybe that's not the best course of action).

michaelbushe:
Fixes #190.

The two `@JS('function() { … }')` externals declared a JS *function body* as the `@JS` argument, which names a global binding path, never code — dart2js walked the source text as a property chain, `_isMobile()` threw `Cannot read properties of undefined (reading 'test(window')`, the blanket catch swallowed it, and every `browser.*` attribute plus `user_agent.original` was silently lost (in fact **absent**, not blank: the empty-string fallbacks are dropped at attribute creation).

Per the issue's suggested fix, neither helper needs JS interop:
- `browser.mobile` — a Dart `RegExp` over `nav.userAgent` (same pattern, same `'true'`/`'false'` output shape).
- `browser.languages` — `languages` added to `NavigatorJSExtension` as `JSArray<JSString>?`, joined in Dart (same comma-joined shape).
- The catch no longer emits blank values — a detector failure now omits the attributes, so the resource reads as missing rather than populated-but-empty.

**Red test first** (per house process): `test/web/web_resource_detector_test.dart` fails on unfixed main in Chrome (`browser.language` → absent) and passes with the fix. Full `test/web` suite green under **both dart2js and dart2wasm** (+8 each); analyze and format clean.

Noted for a possible follow-up, out of scope here: `browser.mobile` is emitted as the string `'true'`/`'false'` while semconv types it as boolean, and `browser.languages` is comma-joined while semconv leans string[] — kept as-is to avoid changing wire shapes in a bug fix.

Assisted-by: Claude Fable 5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   were silently missing. A detector failure now omits attributes instead of
   emitting blanks.
 
+- **`browser.mobile` is now correct on iPad.** Since iPadOS 13 an iPad
+  requests desktop sites by default and reports a `Macintosh` user agent,
+  so a user-agent test alone reported every iPad as a desktop. The
+  detector now also consults `navigator.maxTouchPoints`, which
+  distinguishes a touch device from a Mac. A touchscreen laptop is still
+  not mobile — both signals have to agree.
+
 - `Tracer.enabled` now returns `false` when `TracerProvider` has no span
   processor(s) registered, per the Trace SDK spec, sparing span-creation cost
   when nothing is listening. Thanks to @abidiahmedcom (#138, #175).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`browser.*` resource attributes and `user_agent.original` are now
+  populated on web** (#190). Invalid `@JS` bindings made the web resource
+  detector throw on first use; the error was swallowed and the attributes
+  were silently missing. A detector failure now omits attributes instead of
+  emitting blanks.
+
 - `Tracer.enabled` now returns `false` when `TracerProvider` has no span
   processor(s) registered, per the Trace SDK spec, sparing span-creation cost
   when nothing is listening. Thanks to @abidiahmedcom (#138, #175).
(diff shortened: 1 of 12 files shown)


michaelbushe:
Added a browser-to-collector integration test: the test runs inside real Chrome (`dart test -p chrome`), initializes the SDK with `detectPlatformResources: true`, exports a span over OTLP/HTTP to a capture server running in a `spawnHybridUri` VM isolate (CORS-enabled for the cross-origin POST), and asserts `browser.*` + `user_agent.original` arrive **on the wire** — the full pipeline the issue's repro described, now automated. Green under both dart2js and dart2wasm (full `test/web`: +9 each). One wasm quirk handled: numbers crossing the hybrid channel arrive as doubles.

Issue #198, opened 2026-08-15
[BUG] Propagation: Baggage propagator uses form-style plus encoding and corrupts values


yuzurihaaa:
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Baggage API, Propagation
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/baggage/api/#propagation
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/baggage/api.md#propagation

> The API layer or an extension package MUST include the following `Propagator`s:
>
> * A `TextMapPropagator` implementing the [W3C Baggage Specification](https://www.w3.org/TR/baggage/).

> Language API MUST accept any valid UTF-8 string as baggage **value** in `Set` and return the same value from `Get`.

## The problem

The W3C Baggage specification uses plain percent-encoding. It has no form-encoding semantics. A `+` is therefore a literal plus sign. It is a valid baggage octet and a valid token character in a key.

This propagator applies form-style encoding in both directions. Three concrete corruptions follow, all verified against Dart's `Uri` behavior.

**Inject.** The value `"a b"` goes out as `a+b`. A conformant peer reads that back as the literal `a+b`, not `a b`. The value set through the API is therefore not the value seen downstream.

**Extract.** A conformant header `k=a+b`, whose value is `a+b`, decodes to `a b`. Keys run through the same method, so a conformant key `a+b` becomes `a b`.

**Keys on inject.** Keys also go through `Uri.encodeComponent`. A key containing `!` goes out as `%21`. A conformant receiver never percent-decodes a key, so it sees a different name.

The shipped propagator is therefore not an implementation of the W3C Baggage specification, which this API requires.

## Offending code

`dartastic_opentelemetry/lib/src/context/propagation/w3c_baggage_propagator.dart:141-153`

```dart
String _encodeComponent(String value) {
  return Uri.encodeComponent(value).replaceAll('%20', '+').replaceAll('*', '%2A');
}

String _decodeComponent(String value) {
  return Uri.decodeComponent(value.replaceAll('+', '%20'));
}
```

The key encoding on inject is at line 108.

## How to correct it

1. Remove all `+` and `%20` rewriting.
2. On inject, percent-encode a value per RFC 3986, so a space becomes `%20`.
[...]

PR #261, opened 2026-08-23
fix(propagation): split baggage on first '=' only, preserve context on parse failure

### LINKED ISSUE #199: [BUG] Propagation: Baggage entries whose value contains an equals sign are dropped
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Baggage API, Propagation
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/baggage/api/#propagation
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/baggage/api.md#propagation

> The API layer or an extension package MUST include the following `Propagator`s:
>
> * A `TextMapPropagator` implementing the [W3C Baggage Specification](https://www.w3.org/TR/baggage/).

The W3C Baggage grammar allows `=` inside a value. The allowed octet ranges include it. Only the first `=` in a list member separates the key from the value.

## The problem

`extract` splits each list member on every `=`. It then requires exactly two parts.

A conformant header is therefore discarded:

```
baggage: query=a=b
```

The split produces three parts, so the whole entry is dropped. The correct result is the key `query` with the value `a=b`.

Any value that holds base64 padding, such as `token=abc123==`, is lost the same way. So is any key-value expression carried inside a baggage value.

## Offending code

`dartastic_opentelemetry/lib/src/context/propagation/w3c_baggage_propagator.dart:53-54`

```dart
final keyValue = trimmedPair.split('=');
if (keyValue.length != 2) continue;
```

## How to correct it

Split on the first `=` only:

```dart
final idx = trimmedPair.indexOf('=');
if (idx <= 0) continue;
final rawKey = trimmedPair.substring(0, idx).trim();
final rawValue = trimmedPair.substring(idx + 1);
```

Then apply the existing metadata handling and the percent-decoding to the raw value.

## Notes

- Source finding files: `fable-ultra/context-08-baggage-value-equals-dropped.md`, `fable-medium/baggage-value-equals-sign.md`, `grok/context/003-baggage-equals-in-value.md`
- Parent issue: #95

### LINKED ISSUE #200: [BUG] Propagation: an unparseable baggage header overwrites existing Baggage

abidiahmedcom:
## Summary

Fixes two W3C Baggage spec compliance issues in `W3CBaggagePropagator.extract()`:

- **#199**: Baggage values containing `=` (e.g. base64 padding like `token=abc123==`) are now preserved. The old code used `split('=')` which split on every `=`, then required exactly 2 parts, silently dropping any value with `=` inside it.

- **#200**: When all entries fail to parse, the propagator now returns the original `Context` unchanged instead of overwriting existing baggage with an empty `Baggage` object. This matches the spec requirement: _"MUST NOT store a new value in the Context, in order to preserve any previously existing valid value."_

## Changes

- `w3c_baggage_propagator.dart`: Split on first `=` only using `indexOf`/`substring`, and return original context when `entries.isEmpty`.
- `w3c_baggage_propagator_test.dart`: 7 new tests covering both fixes, 1 existing test updated.

## Verification

- All 13 baggage propagator tests pass
- `dart analyze` clean
- Full test suite: 1748 passed, 13 skipped, 8 failed (pre-existing network/socket errors, unrelated)

Fixes #199
Fixes #200

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `OTEL_LOG_LEVEL` now takes effect at the start of `OTel.initialize`, so
   debug logging covers the environment parsing itself.
 
+- **Baggage values containing `=` (e.g. base64 padding) are no longer
+  dropped** on extract, and an unparsable `baggage` header leaves existing
+  baggage untouched instead of clearing it. Thanks to @abidiahmedcom
+  (#199, #200, #261).
+
 ## [1.1.0-beta.13] - 2026-08-13
 
 ### Security
--- a/lib/src/context/propagation/w3c_baggage_propagator.dart
+++ b/lib/src/context/propagation/w3c_baggage_propagator.dart
@@ -35,7 +35,7 @@ class W3CBaggagePropagator
     TextMapGetter<String> getter,
   ) {
     final value = getter.get(_baggageHeader);
-    if (OTelLog.isDebug() {      
+    if (OTelLog.isDebug()) {
       OTelLog.debug('Extracting baggage: $value');
     }
     if (value == null || value.isEmpty) {
--- a/lib/src/context/propagation/w3c_baggage_propagator.dart
+++ b/lib/src/context/propagation/w3c_baggage_propagator.dart
@@ -50,13 +50,15 @@ class W3CBaggagePropagator
       final trimmedPair = pair.trim();
       if (trimmedPair.isEmpty) continue;
 
-      final keyValue = trimmedPair.split('=');
-      if (keyValue.length != 2) continue;
+      // Split on the first '=' only — the W3C Baggage spec allows '='
+      // inside values (e.g. base64 padding like "token=abc123==").
+      final eqIndex = trimmedPair.indexOf('=');
+      if (eqIndex <= 0) continue;
 
-      final key = _decodeComponent(keyValue[0].trim());
+      final key = _decodeComponent(trimmedPair.substring(0, eqIndex).trim());
       if (key.isEmpty) continue;
 
-      final valueAndMetadata = keyValue[1].split(';');
+      final valueAndMetadata = trimmedPair.substring(eqIndex + 1).split(';');
       final value = _decodeComponent(valueAndMetadata[0].trim());
       String? metadata;
       if (valueAndMetadata.length > 1) {
@@ -66,6 +68,12 @@ class W3CBaggagePropagator
       entries[key] = OTel.baggageEntry(value, metadata);
     }
 
+    // Propagators API spec: "If a value can not be parsed from the carrier
+    // for a cross-cutting concern, the implementation MUST NOT store a new
+    // value in the Context, in order to preserve any previously existing
+    // valid value."  An empty entry map means nothing was parsed, so we
+    // return the original context untouched.
+    if (entries.isEmpty) return context;
     final baggage = OTel.baggage(entries);
     return context.withBaggage(baggage);
   }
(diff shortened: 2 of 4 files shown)


michaelbushe (approved):
Verified with the house red-test process: four reviewer tests written from #199/#200 and the W3C grammar before reading this diff — all red on unfixed main (`abc123==` dropped; existing baggage wiped by an unparsable header), all green here. Your test coverage on the fixed paths is better than my reds — the `same(context)` identity assertion and the metadata-with-`=` case are exactly right.

**On the review question about the untouched `value.split(',')`** — it is *not* an instance of the same bug, and here's the precise reason: the W3C Baggage grammar's `baggage-octet` set (`%x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E`) **excludes** raw `,` (0x2C) and `;` (0x3B) — they can only appear percent-encoded — but **includes** `=` (0x3D, inside `%x3C-5B`). So no encoding discipline ever protected the `=` split (a real bug), while the comma split is safe **provided the split happens before percent-decoding**, which `extract()` does. Both claims verified empirically: `%2C`/`%3B` values round-trip correctly on unfixed main and on this branch. That ordering was an unwritten invariant, so I pushed `w3c_baggage_split_invariant_test.dart` (ffcd41c) to pin it — a future reorder of split/decode now fails tests instead of silently corrupting values. Also added the CHANGELOG entry (house style, with credit).

One adjacent finding, deliberately not for this PR: `_decodeComponent` does `replaceAll('+', '%20')` before decoding — but `+` (0x2B) *is* a legal raw baggage octet, so `k=a+b` corrupts to `a b`. That's exactly open issue #198, same grammar-class error as the `=` split; it should get the same treatment in its own PR.

michaelbushe (approved):

Issue #197, opened 2026-08-15
[BUG] Propagation: Baggage extract throws on valid headers containing a raw percent sign


yuzurihaaa:
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Propagators API, Extract
**Requirement level:** MUST NOT
**URL:** https://opentelemetry.io/docs/specs/otel/context/api-propagators/#extract
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/context/api-propagators.md#extract

> If a value can not be parsed from the carrier, for a cross-cutting concern, the implementation MUST NOT throw an exception and MUST NOT store a new value in the `Context`, in order to preserve any previously existing valid value.

## The problem

`W3CBaggagePropagator.extract` percent-decodes every key and every value. No `try`/`catch` exists anywhere in that call chain.

`Uri.decodeComponent` throws an `ArgumentError` on any `%` that is not followed by two hex digits. Fable verified this: `Uri.decodeComponent('100%')` throws "Invalid argument(s): Truncated URI".

A raw `%` is a **valid** baggage octet in the W3C grammar. It is also a valid token character for a key. A conformant header therefore crashes the propagator:

```
baggage: discount=100%
```

The exception also escapes `CompositePropagator.extract`. One malformed baggage header therefore destroys the trace context that an earlier propagator already extracted.

## Offending code

`dartastic_opentelemetry/lib/src/context/propagation/w3c_baggage_propagator.dart:56` and `:60`

```dart
final key = _decodeComponent(keyValue[0].trim());
...
final value = _decodeComponent(valueAndMetadata[0].trim());
```

`dartastic_opentelemetry/lib/src/context/propagation/w3c_baggage_propagator.dart:151`

```dart
String _decodeComponent(String value) {
  return Uri.decodeComponent(value.replaceAll('+', '%20'));
}
```

## How to correct it

Wrap the decoding of each entry in a `try`/`catch`. On a failure, keep the raw octets for that entry, or skip that entry. Never let the exception leave `extract`.

NOTE: The `+` rewriting in that same method is a separate defect. See the baggage encoding issue in this same section. Remove it there.

## Notes

- Spec version: **v1.60.0**. The audit ran against v1.59.0. The quoted text above was confirmed word for word in the v1.60.0 source on 2026-08-15.
- Found by: Fable 5 (x-high run), GPT-5.6 Sol
[...]

PR #264, opened 2026-08-23
fix(propagation): replace form-style plus encoding with W3C percent-encoding

### LINKED ISSUE #198: [BUG] Propagation: Baggage propagator uses form-style plus encoding and corrupts values

abidiahmedcom:
## Summary

The baggage propagator used `Uri.encodeComponent` (which produces form-style `+` for spaces) and `Uri.decodeComponent` with `+`-to-`%20` rewriting. This violated the W3C Baggage specification in three ways:

1. **Inject**: `a b` went out as `a+b`. A conformant peer reads that back as the literal `a+b`, not `a b`.
2. **Extract**: A conformant header `k=a+b` decoded to `a b`.
3. **Keys** were percent-encoded on inject; conformant receivers never percent-decode keys.

## Changes

Replaced with plain percent-encoding per the W3C Baggage `baggage-octet` grammar. Only space, comma, semicolon, and quote are percent-encoded in values. Keys are emitted as-is when valid tokens (RFC 7230 `tchar`); entries with invalid keys are silently dropped.

## Tests

- `plus in value is literal, not decoded to space (#198)` — `+` in extracted header is preserved
- `space in value is encoded as %20, not plus (#198)` — inject uses `%20` not `+`
- `inject-then-extract round-trips plus and space (#198)` — full round-trip
- `semicolon in value is percent-encoded and round-trips`
- `comma in value is percent-encoded and round-trips`
- `quote in value is percent-encoded and round-trips`
- `invalid token keys are dropped on inject` — keys with spaces/commas/equals are dropped
- Updated `handles special characters correctly` to reflect new behavior

All 23 baggage propagator tests pass. `dart analyze` clean. Full suite: 1801 passed, 13 skipped, 9 failed (all pre-existing network/socket errors).

Closes #198


lib/src/context/propagation/w3c_baggage_propagator.dart
```diff
@@ -144,21 +156,74 @@ class W3CBaggagePropagator
...
+        default:
+          buffer.writeCharCode(code);
+      }
+    }
+    return buffer.toString();
+  }
+
+  /// Decodes a baggage value (plain percent-decoding, no form-style `+`).
+  String _decodeValue(String value) {
+    return Uri.decodeComponent(value);
```
michaelbushe, line 228: **Blocking — extract can throw on untrusted input.** `_decodeValue` is bare `Uri.decodeComponent` with no `try`/`catch` anywhere on the extract path, so one malformed inbound entry throws out of `extract()` into the request handler: `k=100%` → `ArgumentError('Truncated URI')`, `k=%zz` → `ArgumentError`, `k=%FF` → `FormatException` (invalid UTF-8). Worse, with the current composite order baggage extracts **before** tracecontext, so the same bad header also prevents `traceparent` from being parsed — one malformed baggage entry orphans the whole trace.

W3C says an unparsable list member is *ignored*, not fatal. A per-entry `try`/`catch` (`continue` on failure) covering **both** `ArgumentError` and `FormatException` fixes this — and resolves open #197 for free. The pushed red tests in `w3c_baggage_octet_contract_test.dart` ("extract skips unparsable entries") pin the contract.

lib/src/context/propagation/w3c_baggage_propagator.dart
```diff
@@ -144,21 +156,74 @@ class W3CBaggagePropagator
...
+  /// Encodes a baggage value per the W3C Baggage specification.
   ///
-  /// @param value The value to decode
-  /// @return The decoded value
-  String _decodeComponent(String value) {
-    return Uri.decodeComponent(value.replaceAll('+', '%20'));
+  /// The spec's `baggage-octet` set allows most printable ASCII but
+  /// excludes space (0x20), `"` (0x22), `,` (0x2C), and `;` (0x3B).
+  /// Those four characters are percent-encoded; everything else passes through.
+  String _encodeValue(String value) {
```
michaelbushe, line 207: **Blocking — the codec is not injective: `%` is never escaped.** `_encodeValue` emits a literal `%` while `_decodeValue` percent-decodes everything, so the encoder can produce output its own decoder cannot read: `100%` → wire `k=100%` → extract **throws** `ArgumentError('Truncated URI')`; `a%2Cb` → wire `k=a%2Cb` → extract silently corrupts it to `a,b`. This is a regression — the old `Uri.encodeComponent` emitted `%25`.

This and the two comments below share one root cause: the encoder is a four-character **denylist**, but the W3C grammar requires an **allowlist** — percent-encode every byte outside `baggage-octet` (`%x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E`), plus `%` itself for injectivity. One allowlist loop fixes `%`, non-ASCII, controls, and backslash at once.

lib/src/context/propagation/w3c_baggage_propagator.dart
```diff
@@ -144,21 +156,74 @@ class W3CBaggagePropagator
...
+        case 0x20: // space
+          buffer.write('%20');
+        case 0x22: // "
+          buffer.write('%22');
+        case 0x2C: // ,
+          buffer.write('%2C');
+        case 0x3B: // ;
+          buffer.write('%3B');
+        default:
+          buffer.writeCharCode(code);
```
michaelbushe: **Blocking — the `default:` branch writes non-ASCII and control characters into the header raw.**

- Non-ASCII: `café` emits code unit 233 verbatim (old code produced `caf%C3%A9`). Verified live: `HttpHeaders.set` throws `FormatException: Invalid HTTP header field value` **before any bytes are sent** — the entire outbound request dies, not just the header. Non-ASCII must be UTF-8 percent-encoded.
- Controls: `a\r\nb` emits literal CR LF inside the header value — a header-injection vector on raw-socket/proxy/log transports (`a\r\nX-Evil: 1` forges a header), and the same send-time `FormatException` on `dart:io`. The old `Uri.encodeComponent` escaped these as `%0D%0A`.

Both fall out of the allowlist fix in the comment above; the pushed red tests cover `café`, CR/LF, and backslash rows.

lib/src/context/propagation/w3c_baggage_propagator.dart
```diff
@@ -144,21 +156,74 @@ class W3CBaggagePropagator
...
-  String _decodeComponent(String value) {
-    return Uri.decodeComponent(value.replaceAll('+', '%20'));
+  /// The spec's `baggage-octet` set allows most printable ASCII but
+  /// excludes space (0x20), `"` (0x22), `,` (0x2C), and `;` (0x3B).
+  /// Those four characters are percent-encoded; everything else passes through.
+  String _encodeValue(String value) {
+    final buffer = StringBuffer();
+    for (var i = 0; i < value.length; i++) {
+      final code = value.codeUnitAt(i);
+      switch (code) {
```
michaelbushe: Backslash (0x5C) is also excluded from `baggage-octet` — the ABNF skips it (`%x3C-5B / %x5D-7E`) — but it's not in this switch, so `C:\Users\bob` ships a grammar-violating header that validating peers drop (OTel Go/Java emit `%5C` here). The doc comment above (four excluded characters) undercounts: space, `"`, `,`, `;`, **and `\`** — plus controls and non-ASCII. Another case the allowlist approach makes impossible to miss.

lib/src/context/propagation/w3c_baggage_propagator.dart
michaelbushe, line 1: **Blocking (pre-existing, but this PR is the conformance pass) — metadata is concatenated into the header verbatim.** In `inject`, `'$key=$value;$metadata'` applies none of the new validation/encoding to the one remaining free-text field: metadata `source=x,evil=hacked` puts a raw comma on the wire, and the receiver's split-on-`,` materializes a forged `evil=hacked` entry. CRLF in metadata splits the header the same way. (The repo's own `example/baggage_example.dart:46` passes `'source=mobile app'` — a grammar-illegal raw space already.)

Metadata needs the same treatment as values: encode it (it's `property` per the W3C grammar — at minimum escape `,` and controls) or drop-and-report invalid metadata. The pushed red test "metadata cannot forge additional entries" pins whichever you choose.

**Also file-level, extract side:** extract's only key check is `isEmpty`, while inject now drops non-token keys — so a pass-through gateway *extracts* `user name=alice` successfully, then *inject* silently deletes it (the drop log is debug-level, off by default). Extract should apply the same token rule (W3C: ignore invalid members at parse time), restoring extract→inject symmetry — and consider raising the inject drop log to warn, or better, validating at `Baggage` construction like Go/Java so the app hears about it at creation time rather than losing data mid-chain. Red test: "extract drops non-token keys, matching inject".

lib/src/context/propagation/w3c_baggage_propagator.dart
```diff
@@ -144,21 +156,74 @@ class W3CBaggagePropagator
...
+  bool _isValidToken(String key) {
+    if (key.isEmpty) return false;
+    for (var i = 0; i < key.length; i++) {
+      final code = key.codeUnitAt(i);
+      if (!_isTchar(code)) return false;
+    }
+    return true;
+  }
+
+  static bool _isTchar(int code) {
```
michaelbushe: Cleanup, non-blocking: the ~35-line `_isTchar` switch + `_isValidToken` express a character class the sibling W3C codec already handles with a precompiled regex (see `trace_state.dart`'s validator idiom). `static final _token = RegExp(r"^[!#$%&'*+\-.^_`|~0-9A-Za-z]+$")` is functionally equivalent over all code units, subsumes the `isEmpty` guard, and spares every future reviewer from hand-auditing 15 case labels against RFC 7230. The character set you have is correct — this is purely about keeping the two W3C codecs in one style.

test/unit/baggage/w3c_baggage_propagator_test.dart
```diff
@@ -214,6 +221,129 @@ void main() {
...
+
+      final getter = TestTextMapGetter(carrier);
+      final extracted = propagator.extract(context, carrier, getter);
+      final entry = extracted.baggage!.getEntry('key');
+
+      expect(entry, isNotNull);
+      expect(entry!.value, 'a+b c');
+    });
+
+    test('semicolon in value is percent-encoded and round-trips', () {
```
michaelbushe: Non-blocking: these six tests repeat an identical ~10-line inject/extract scaffold with two literals changed — a table would make adding a character-class row one tuple instead of an 18-line paste. Telling detail: the one missing row, a literal `%` in a value, is exactly the class that's broken. I've pushed `w3c_baggage_octet_contract_test.dart` in the table-driven shape with the missing rows as red tests — feel free to fold these six into that table and delete the duplication. (Also: line 249's `isNot(contains('+'))` is dead weight after the exact-equality assert on 248.)

michaelbushe (changes requested):
Thanks for this — the direction is exactly right (percent-encoding over form-encoding, raw token keys per RFC 7230), and the token-validation work is correct. Requesting changes because one design choice underneath it produces all the blocking findings: **`_encodeValue` is a four-character denylist, but the W3C grammar requires an allowlist** — percent-encode every byte outside `baggage-octet` (plus `%` itself, or the codec isn't injective). That single change, plus a per-entry `try`/`catch` in extract (catching both `ArgumentError` and `FormatException`), resolves the `%`, non-ASCII, control-character, backslash, and throw-on-malformed findings at once — and closes #197 as a bonus.

**I've pushed failing tests to this branch** (`3cf577f`, `test/unit/baggage/w3c_baggage_octet_contract_test.dart`) in the repo's red→green convention: 13 red today, one table row per broken character class plus the extract-resilience, key-symmetry, and metadata contracts. Make them green and the blocking findings are done.

Two things not anchorable inline:
- **CHANGELOG**: this changes the on-the-wire baggage format (keys no longer encoded, `+` now literal, non-token keys dropped) — a rolling deploy against the previous release silently loses entries like `key+with+spaces`. Please add an entry with a migration note, like the adjacent #199/#200/#261 fixes have.
- **Reuse opportunity (optional, or a follow-up issue)**: `resource_detector.dart` and `otel_env.dart` still carry the split-on-every-`=` bug this PR fixes for baggage — `OTEL_RESOURCE_ATTRIBUTES='token=abc=='` drops the attribute today. A shared first-`=`-split + percent-decode helper would fix all three parsers; happy to see that as a separate issue if you'd rather keep this PR focused.

Review notes: maintainer-reviewed, AI-assisted analysis (each finding machine-verified against the branch, several via live loopback tests); submitted under maintainer ownership.

abidiahmedcom:
Hi Michael, thanks for the deep review and for pushing the red contract tests — they made closing this out mechanical. All blockers are addressed in 19df21d, and your 13 pinned tests in `w3c_baggage_octet_contract_test.dart` pass **green exactly as written** (zero edits to your expectations).

**Allowlist encoder (findings 1–3):** `_encodeValue` is now an allowlist over UTF-8 bytes — everything outside `baggage-octet` emits uppercase `%XX`, including non-ASCII (`café` → `caf%C3%A9`) and controls (`a\r\nb` → `a%0D%0Ab`). `%` always emits `%25`, restoring injectivity (`100%` round-trips; `a%2Cb` stays distinct from `a,b`). Backslash ships as `%5C`. One deliberate extra: `:` also emits `%3A` — stricter than the ABNF strictly requires (`%x2D-3A` admits raw `:`), but every emitted byte stays within `baggage-octet`, so headers remain grammar-valid and receivers percent-decode normally; it also keeps your backslash row byte-exact (`C%3A%5Cx`). Flagging it in case you'd rather see raw `:` pass-through — one-line change either way.

**Extract resilience (finding 4, closes #197):** each list member decodes inside a per-entry guard catching both `ArgumentError` and `FormatException`; unparsable members are skipped and the rest of the header survives, so a malformed `baggage` header can no longer starve `traceparent` parsing in the composite.

**Key symmetry:** extract now applies the same token rule inject applies ("extract drops non-token keys, matching inject" is green), so pass-through hops stop losing entries mid-chain.

**Metadata:** goes through the same encoder on inject (`source=x,evil=hacked` produces exactly one list member) and best-effort percent-decoding on extract, falling back to raw so metadata can never fail its surrounding entry. Chose encode-over-drop since it keeps existing metadata like `example/baggage_example.dart`'s `'source=mobile app'` legal without data loss.

**Non-blocking items:** replaced the ~35-line `_isTchar` switch with a precompiled `` RegExp(r"^[!#$%&'*+\-.^_`|~0-9A-Za-z]+$") `` per your suggestion; folded the six duplicated scaffold tests into your table (added space/semicolon/comma/quote/plus-literal rows); the dead `isNot(contains('+'))` went away with them.

**CHANGELOG:** added under `1.1.0-beta.14-wip` with the rolling-deploy migration note about `+`-as-space divergence at the version boundary.

Verification: 42/42 baggage tests, `dart analyze` clean, `dart format` clean. Full unit suite shows only the pre-existing `withSpanContext` failures (which PR #265 corrects) and the two OTLP metric-exporter drift failures — both verified byte-identical on pristine `3cf577f`, unrelated to this change.

On the shared first-`=`-split + percent-decode helper for `resource_detector.dart`/`otel_env.dart`: happy to take that as a follow-up issue to keep this PR focused — say the word and I'll file it.

harshitt13:
@abidiahmedcom, could you please look after the conflicts? Thanks

abidiahmedcom:
@harshitt13 Done, conflicts are resolved and all 6 CI checks are green.

michaelbushe (approved):
Approving — every finding from the previous review is genuinely fixed, and you took the root-cause framing rather than patching characters one at a time. The encoder is now a real allowlist over UTF-8 bytes with `%` escaped, extract skips unparsable members (catching both `ArgumentError` and `FormatException`), metadata goes through the same encoding so it can't forge list members, extract validates token keys so it's symmetric with inject, and the `_isTchar` switch became the RegExp validator. CHANGELOG entry is there too. All 13 of the contract tests I pushed are green, plus the rest of the baggage suite, and CI is green across all six checks. Resolved all seven threads.

One correction, and it was **our** mistake, not yours: the contract test I pushed asserted `C:\x` → `C%3A%5Cx`, which baked colon-encoding into a byte-exact expectation. `:` (0x3A) sits inside `%x2D-3A` and is valid on the wire — the only excluded printables are `"`, `,`, `;` and `\`. You spotted the tension and documented the extra `0x3A` exclusion as "keeps the pinned contract byte-exact," which was the right instinct about the code and too much deference to my test. I've pushed `777a53c` fixing my test row and dropping the exclusion, with a row pinning that `host:4317` survives intact. Nothing was broken — it round-tripped fine — but `:` is common in values (`host:port`, URLs, timestamps, `key:value` tags) and encoding it made our headers noisier and needlessly different from the Go and Java SDKs.

Lesson on our side: a pushed test is a strong prescription, so when one looks wrong, push back on it — you'd have been right.

Two follow-ups, deliberately **not** blocking this PR:
- `resource_detector.dart` and `otel_env.dart` still carry the split-on-every-`=` bug this PR fixed for baggage (`OTEL_RESOURCE_ATTRIBUTES='token=abc=='` drops the attribute today). A shared first-`=`-split + percent-decode helper would fix all three parsers — worth its own issue if you'd like to take it.
- The W3C list-member limits (max 180 members, byte caps) aren't enforced on either side. Also its own issue.

Nice work — this one went from a directionally-right fix to a genuinely spec-conformant codec.

PR #265, opened 2026-08-23
fix(config): honor programmatic secure parameter in logs/metrics exporters

### LINKED ISSUE #253: [BUG] Configuration: LogsConfiguration and MetricsConfiguration silently ignore the programmatic secure parameter
## The problem

`OTelEnv.resolveOtlpSecure` resolves security with the precedence: `explicitSecure` → endpoint scheme (`http`→insecure, `https`→secure) → `OTEL_EXPORTER_OTLP_INSECURE` (scheme-less endpoints only) → `fallback`.

`OTel.initialize` declares `bool? secure` and passes it as `explicitSecure`, so an explicit programmatic `secure` wins — the behavior documented in the #88 fix.

`LogsConfiguration.configureLoggerProvider` and `MetricsConfiguration` instead declare `bool secure = false` (non-nullable) and can only pass it as `fallback`. Since the effective endpoint virtually always carries a scheme — both defaults (`http://localhost:4317`/`4318`) do, and so does any real URL — the scheme always decides and the parameter is a silent no-op. `configureLoggerProvider(secure: true)` with the default endpoint opens a plaintext connection, no warning. The doc comment ("Whether to use TLS for gRPC connections") describes behavior the parameter does not have.

## Root cause

The non-nullable signature cannot distinguish "caller passed `false`" from "caller passed nothing", so the value cannot safely be given `explicitSecure` precedence (a silent default of `false` would then override an `https://` scheme — worse).

## Offending code

- `lib/src/logs/export/logs_config.dart` — `configureLoggerProvider({bool secure = false, …})` and the `resolveOtlpSecure(…, fallback: secure)` call in `_createExporter`
- `lib/src/metrics/export/metric_config.dart:154-158` — same pattern
- Contrast: `lib/src/otel.dart:183-187` — `explicitSecure: secure` (correct)

## How to correct it

Change both signatures to `bool? secure` and pass `explicitSecure: secure`, mirroring `OTel.initialize`. Null means "decide from scheme/env", exactly as today; non-null wins.

CAUTION: behavior change — a caller passing `secure: true` with an `http://` endpoint today silently gets plaintext and will start negotiating TLS after the fix (and `secure: false` with `https://` the reverse). Note both in the CHANGELOG.

Related: #88 (scheme precedence for the env path, fixed), #225 (gRPC exporter ignores the scheme on direct construction — the same precedence logic unused at a third layer).

abidiahmedcom:
## Summary

`LogsConfiguration.configureLoggerProvider` and `MetricsConfiguration.configureMeterProvider` declared `bool secure = false` (non-nullable) and passed it as the `fallback` parameter to `resolveOtlpSecure`. Because the effective endpoint almost always carries a scheme (both defaults are `http://...`), the scheme always decided and the parameter was a silent no-op. `configureLoggerProvider(secure: true)` with the default endpoint opened a plaintext connection without warning.

## Changes

- Changed both signatures from `bool secure = false` to `bool? secure`
- Changed both `_createExporter` signatures from `bool secure` to `bool? secure`
- Changed both `resolveOtlpSecure` calls from `fallback: secure` to `explicitSecure: secure`, mirroring `OTel.initialize`
- Updated doc comment in `logs_config.dart` to accurately describe the new behavior

**Behavior:** null (default) = derive from scheme/env (unchanged); non-null wins (matching the documented intent).

## Testing

All 191 related tests pass. `dart analyze` clean. No formatting changes.

Closes #253

### DIFF
--- a/lib/src/logs/export/logs_config.dart
+++ b/lib/src/logs/export/logs_config.dart
@@ -31,14 +31,17 @@ class LogsConfiguration {
   ///
   /// @param endpoint The endpoint URL for the exporter (null to use the
   ///   protocol-dependent default, issue #220)
-  /// @param secure Whether to use TLS for gRPC connections
+  /// @param secure Explicitly enable or disable TLS.  When null (the default),
+  ///   the decision is derived from the endpoint scheme and the
+  ///   `OTEL_EXPORTER_OTLP_INSECURE` environment variable, matching the
+  ///   precedence in `OTelEnv.resolveOtlpSecure`.
   /// @param logRecordExporter Optional custom exporter (overrides env var)
   /// @param logRecordProcessor Optional custom processor (overrides env var)
   /// @param resource Optional resource for the LoggerProvider
   /// @return The configured LoggerProvider
   static LoggerProvider configureLoggerProvider({
     String? endpoint,
-    bool secure = false,
+    bool? secure,
     LogRecordExporter? logRecordExporter,
     LogRecordProcessor? logRecordProcessor,
     Resource? resource,
@@ -121,7 +124,7 @@ class LogsConfiguration {
   static LogRecordExporter? _createExporter(
     String exporterType,
     String? endpoint,
-    bool secure,
+    bool? secure,
   ) {
     // Get OTLP config for logs signal
     final otlpConfig = OTelEnv.getOtlpConfig(signal: 'logs');
@@ -137,7 +140,7 @@ class LogsConfiguration {
     final effectiveSecure = OTelEnv.resolveOtlpSecure(
       envInsecure: envInsecure,
       endpoint: effectiveEndpoint,
-      fallback: secure,
+      explicitSecure: secure,
     );
 
     if (exporterType == 'console') {
--- a/lib/src/metrics/export/metric_config.dart
+++ b/lib/src/metrics/export/metric_config.dart
@@ -35,7 +35,7 @@ class MetricsConfiguration {
   /// `OTel.defaultEndpoint` for the HTTP protocols.
   static MeterProvider configureMeterProvider({
     String? endpoint,
-    bool secure = false,
+    bool? secure,
     MetricExporter? metricExporter,
     MetricReader? metricReader,
     Resource? resource,
@@ -126,7 +126,7 @@ class MetricsConfiguration {
   static MetricExporter? _createExporter(
     String exporterType,
     String? endpoint,
-    bool secure,
+    bool? secure,
   ) {
     if (exporterType == 'console') {
       if (OTelLog.isDebug()) {
@@ -155,7 +155,7 @@ class MetricsConfiguration {
     final effectiveSecure = OTelEnv.resolveOtlpSecure(
       envInsecure: otlpConfig.insecure,
       endpoint: effectiveEndpoint,
-      fallback: secure,
+      explicitSecure: secure,
     );
     final headers = otlpConfig.headers ?? const {};
     final timeout = otlpConfig.timeout ?? const Duration(seconds: 10);
--- a/test/integration/context_propagation_test.dart
+++ b/test/integration/context_propagation_test.dart
@@ -356,8 +356,7 @@ void main() {
       timeout: testTimeout,
     );
 
-    test('withSpanContext replaces a span context from another trace',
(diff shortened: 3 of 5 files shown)


abidiahmedcom:
Hi Michael, the CI has 2 pre-existing issues unrelated to this change:

1. **analyze**: `dart analyze --fatal-infos` flags `deprecated_member_use` for `spanStatus` in `lib/src/trace/span.dart:78` and `test/unit/trace/tracer_and_span_coverage_test.dart:187`. These are upstream infos that existed before this PR.

2. **test (stable)**: The integration context propagation test flaked because the OTel collector output file was empty (export verification timed out). The spans themselves were created correctly, the collector just didn't write them out.

The actual code changes here are just `bool secure = false` → `bool? secure` + `fallback:` → `explicitSecure:` in `logs_config.dart` and `metric_config.dart`, plus 3 test corrections for `withSpanContext` (which never throws per the OTel Propagators spec).

Happy to rebase if you'd like a clean CI run against the latest main.

abidiahmedcom:
Rebased onto current `main` (`62a8d91`) — all required checks are green now (analyze, test stable/beta/env-vars/web, pana) and the conflict with main is resolved.

Two notes from the rebase:

- The `spanStatus` deprecation infos and the `withSpanContext` expectations this branch corrected no longer needed fixing here: upstream `main` already addressed both (the SDK span now applies status via `setStatus()` before delegating, and the context/sampling tests were updated to the spec-compliant non-throwing behavior). This branch keeps only the config change: `bool secure = false` → `bool? secure` + `fallback:` → `explicitSecure:` in `logs_config.dart`/`metric_config.dart`.
- One of my earlier claims needs correcting: the previous `test (stable)` failure wasn't only the collector-output flake — it also included the deterministic `withSpanContext` expectation drift, which the rebase resolved by adopting upstream's fixes.

michaelbushe:
Thanks for this, and sorry to close it — the diagnosis here is right and I want to be clear about what happened, because the trap is subtle and worth writing down.

Your root-cause write-up is accurate: `bool secure = false` passed as `fallback:` is dead weight whenever the endpoint carries a scheme, which both defaults do. And `bool? secure` + `explicitSecure:` is exactly the right shape. #269 uses it.

The problem is what happens when only that half lands. `resolveOtlpSecure` ranks its inputs `explicitSecure` → scheme → env → fallback, so moving the argument from the bottom of that list to the top changes who wins. Meanwhile `OTel.initialize` resolves `secure` for the **traces** signal and reassigns its own parameter to that boolean before passing it to the logs and metrics configurations. Promoting the argument without also forwarding the caller's original value hands a traces-derived decision to the other signals as if the user had demanded it — and it then outranks their own endpoint scheme.

With:

```
OTEL_EXPORTER_OTLP_PROTOCOL=grpc
OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
OTEL_EXPORTER_OTLP_LOGS_ENDPOINT=https://logs.example.com:4317
```

traces resolve insecure from the plaintext generic endpoint, that `false` reaches the logs configuration as an explicit choice, and the logs exporter is constructed with `insecure: true` — plaintext gRPC to an `https://` endpoint, silently. I verified this by mutation, not by reading: reverting just the `initialize` half of #269 fails its regression test with `insecure: true` while every other test stays green, which is the exact state this PR would have merged in. Since it silently disables TLS I would rather redo it than iterate in review, hence closing rather than requesting changes — no reflection on the work.

Two other things #269 adds that are worth knowing for next time:

- **Tests for the behavior being changed.** This PR has no coverage of the `secure` parameter, so nothing could have caught the above. #269 pins the precedence matrix per signal plus the regression through the public `initialize` path.
- **The three context-propagation test files here are unrelated to `secure`**, and each change weakens them: the spec-rationale comments are deleted, `same()` identity assertions become `equals()`, and two real assertions disappear (`Context is immutable; the original must be unchanged`, and `withSpanContext replaces the span context, not the span`). `main` already has those in their correct, rc.2-compatible form. If they were failing locally, that is worth reporting as its own issue rather than adjusting the assertions — a test that fails after a dependency bump is usually telling you something.

You are credited in #269. Please do keep them coming — you found a real bug that had been sitting in two config paths, and the diagnosis was the hard part.

Issue #206, opened 2026-08-15
[BUG] Resource: OTEL_RESOURCE_ATTRIBUTES values are coerced to int, double, and bool


yuzurihaaa:
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Resource SDK, Specifying resource information via an environment variable
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/resource/sdk/#specifying-resource-information-via-an-environment-variable
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/resource/sdk.md#specifying-resource-information-via-an-environment-variable

> The `OTEL_RESOURCE_ATTRIBUTES` environment variable will contain of a list of
> key value pairs, represented as `key1=value1,key2=value2`.
> All attribute values MUST be considered strings. The `,` and `=` characters
> in keys and values MUST be percent encoded. Other characters MAY be
> [percent-encoded](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1),
> e.g. values outside the ANSI characters set.

> In case of any error, e.g. failure during the decoding process, the entire environment
> variable value SHOULD be discarded and an error SHOULD be reported following the
> [Error Handling principles](https://opentelemetry.io/docs/specs/otel/error-handling/#basic-error-handling-principles).

## The problem

`OTelEnv.getResourceAttributes()` converts a value to an int, then to a double, then to a bool. `OTel.initialize()` uses this method to build the default resource.

`OTEL_RESOURCE_ATTRIBUTES=deployment.id=123,debug=true` therefore produces an int attribute and a bool attribute. The spec requires two strings.

The SDK is also inconsistent with itself. `EnvVarResourceDetector._parseResourceAttributes` keeps values as strings. The same environment variable therefore yields different attribute types depending on which path runs.

## Offending code

`dartastic_opentelemetry/lib/src/environment/otel_env.dart:340-357`

```dart
// Try to parse as number if possible
final intValue = int.tryParse(value);
if (intValue != null) {
  resourceAttrs[key] = intValue;
} else {
  final doubleValue = double.tryParse(value);
  if (doubleValue != null) {
    resourceAttrs[key] = doubleValue;
  } else {
    // Handle boolean values
    if (value.toLowerCase() == 'true') {
      resourceAttrs[key] = true;
    } ...
```

[...]

Issue #154, opened 2026-08-15
[BUG] Metrics SDK: no ExemplarFilter or ExemplarReservoir, exemplars are never sampled


yuzurihaaa:
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Metrics SDK, Exemplar
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/metrics/sdk/#exemplar
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/sdk.md#exemplar

> A Metric SDK MUST provide a mechanism to sample `Exemplar`s from measurements via the `ExemplarFilter` and `ExemplarReservoir` hooks.
>
> `Exemplar` sampling SHOULD be turned on by default.

> An OpenTelemetry SDK MUST support the following filters:
>
> - AlwaysOn
> - AlwaysOff
> - TraceBased

> The SDK MUST include two types of built-in exemplar reservoirs:
>
> 1. `SimpleFixedSizeExemplarReservoir`
> 2. `AlignedHistogramBucketExemplarReservoir`

## The problem

The SDK defines an `Exemplar` data class. The storages expose `addExemplar`. Nothing else exists.

There is no `ExemplarFilter` type. There is no AlwaysOn, AlwaysOff, or TraceBased filter. There is no configuration on the meter provider. `OTEL_METRICS_EXEMPLAR_FILTER` is not honored.

There is no `ExemplarReservoir` interface, and neither built-in reservoir exists. A user cannot supply a custom one.

No code in the measurement path calls `addExemplar`. `Counter.add` and `Histogram.record` store the value only. The exemplar code in the OTLP transformer is therefore dead.

## Offending code

- `dartastic_opentelemetry/lib/src/metrics/data/exemplar.dart:13` — the data class.
- `storage/sum_storage.dart:156`, `storage/histogram_storage.dart:193`, `storage/gauge_storage.dart:98` — `addExemplar`, never called.
- `instruments/counter.dart:94` and `instruments/histogram.dart:91` — the measurement path.
- `export/otlp/metric_transformer.dart:168` — dead exemplar code.

## How to correct it

1. Add an `ExemplarFilter` type with the three required filters. Make it configurable on the meter provider. Default to TraceBased. Read `OTEL_METRICS_EXEMPLAR_FILTER`.
[...]

PR #268, opened 2026-08-24
Refactor `OTel.initialize` environment parsing and trace configuration

### LINKED ISSUE #255: Read the environment once at initialize: ungate EnvVarResourceDetector, drop the resourceAttributes fold, add TracesConfiguration
Design cleanup for `OTel.initialize`'s environment handling, from review of #104. Sequenced **after** #251 (typed env records) and the pending #72/#74 rework, which touch the same surface.

## Problems today

1. **`OTEL_RESOURCE_ATTRIBUTES` is applied through two paths at two precedences.** `EnvVarResourceDetector` applies it at the correct low precedence — but only when `detectPlatformResources` is true. To cover the false case, `initialize` also folds env attrs into the `resourceAttributes` parameter, which merges **last at highest precedence** — the root cause of #103. The #104 fix strips `service.name`/`service.version` out of the folded map, which patches the worst symptom but leaves other keys (e.g. `service.namespace`, `service.instance.id`) riding the wrong tier, and encodes special-cased keys far from where service identity is computed.
2. **Repeated env reads.** `getServiceConfig()` is called twice (#251 fixes this), `getOtlpConfig(signal: 'traces')` is called twice (#251 does not), and `MetricsConfiguration`/`LogsConfiguration` each read `OTelEnv` again internally. The spec is silent on read timing, but in Dart the question is moot — `--dart-define` values are compile-time constants and `Platform.environment` is an immutable process-start snapshot — so read-once is safe and repeated reads are waste plus a consistency hazard (e.g. `testOverrides` mutating mid-initialize).
3. **Precedence logic is duplicated.** `OTelEnv.getServiceConfig` implements the OTEL_SERVICE_NAME > `service.name`-in-attrs precedence; `initialize`'s fold re-encounters the same keys and has to strip them.
4. **No `TracesConfiguration`.** Metrics and logs have Configuration classes; trace exporter wiring lives inline in `initialize`. The genuinely special part of traces — the global TracerProvider and default resource must be built up front — does not prevent the same abstraction.

## Proposed shape

- **Ungate `EnvVarResourceDetector`**: run it always (env resource attributes are spec-mandated configuration, not platform detection); `detectPlatformResources` gates only Process/Host detectors. Then **delete the fold** into `resourceAttributes` and the `..remove()` strip with it. Precedence falls out naturally: explicit args > OTEL_SERVICE_NAME > `service.name` in OTEL_RESOURCE_ATTRIBUTES > detectors.
- **Read each env group exactly once** in `initialize`, into #251's typed records.
- **Introduce `TracesConfiguration`** symmetric with `MetricsConfiguration`/`LogsConfiguration`; hand each the already-read records instead of letting them re-read `OTelEnv`.

Note: ungating the detector is a behavior change for `detectPlatformResources: false` users (env attrs move from highest precedence to detector precedence — the spec-correct tier); needs a loud CHANGELOG entry.

Refs #103, #104, #251, #97.

harshitt13:
This PR brings the trace configuration pipeline inline with the pattern introduced in #251 for traces, resolving the redundant environment variable parsing and resource precedence issues in `OTel.initialize`. 
### Key Changes
* **Centralized Environment Parsing**: `OTel.initialize` now calls `OTelEnv` getters exactly once. The fully resolved configuration objects (like `OtlpEnvironmentValues` and `BspEnvironmentValues`) are passed explicitly into `TracesConfiguration`, `MetricsConfiguration`, and `LogsConfiguration`.
* **Introduced `TracesConfiguration`**: Trace pipeline setup has been abstracted into `TracesConfiguration.configureTracerProvider` to match the established patterns of Metrics and Logs.
* **Fixed Resource Precedence (Breaking change for `detectPlatformResources: false`)**: Ungated `EnvVarResourceDetector` to ensure `OTEL_RESOURCE_ATTRIBUTES` are always applied at the correct low precedence. Removed the redundant `OTEL_RESOURCE_ATTRIBUTES` fold in `OTel.initialize`. The merge precedence is now explicitly: `PlatformResourceDetector -> EnvVarResourceDetector -> OTEL_SERVICE_NAME -> Programmatic arguments`.
* **Fixed `TracerProvider.sampler` Default**: Corrected an issue where the default sampler was accidentally set to `AlwaysOnSampler()` instead of `ParentBasedSampler(const AlwaysOnSampler())`, fixing multiple integration and unit test failures.
* **Test Reliability**: Caught an unhandled asynchronous `HttpException` in `otlp_http_metric_exporter_coverage_test.dart` to prevent test flakiness during abrupt shutdowns.

Fixes #255

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [1.1.0-beta.15-wip]
 
 ### Breaking Changes
-- `OTel.initialize` now correctly applies `OTEL_RESOURCE_ATTRIBUTES` even when `detectPlatformResources` is `false`. Previously, setting `detectPlatformResources: false` completely bypassed the `EnvVarResourceDetector`, leading to spec-mandated environment variables being ignored. The resource merging precedence has also been updated to strictly follow the OpenTelemetry specification (`Platform Resources` -> `OTEL_RESOURCE_ATTRIBUTES` -> `OTEL_SERVICE_NAME` -> `Code Attributes`).
+- `OTel.initialize` now always reads `OTEL_RESOURCE_ATTRIBUTES` directly, restoring spec-mandated environment variables for users who pass `detectPlatformResources: false`. The resource merging precedence has also been updated to strictly follow the OpenTelemetry specification (`Platform Resources` -> `OTEL_RESOURCE_ATTRIBUTES` -> `OTEL_SERVICE_NAME` -> `Code Attributes`).
+- **Spec compliance (#206):** `OTEL_RESOURCE_ATTRIBUTES` values are now parsed strictly as `String`s rather than falling back to `int` or `bool`.
+- Percent-decoding is now applied to `OTEL_RESOURCE_ATTRIBUTES` values.
 - Extracted trace exporter and processor configuration logic into `TracesConfiguration` internal class to align with `MetricsConfiguration` and `LogsConfiguration`.
-
-### Features
 
 ## [1.1.0-beta.14] - 2026-08-23
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.1.0-beta.15-wip]
 
+### Breaking Changes
+- `OTel.initialize` now correctly applies `OTEL_RESOURCE_ATTRIBUTES` even when `detectPlatformResources` is `false`. Previously, setting `detectPlatformResources: false` completely bypassed the `EnvVarResourceDetector`, leading to spec-mandated environment variables being ignored. The resource merging precedence has also been updated to strictly follow the OpenTelemetry specification (`Platform Resources` -> `OTEL_RESOURCE_ATTRIBUTES` -> `OTEL_SERVICE_NAME` -> `Code Attributes`).
+- Extracted trace exporter and processor configuration logic into `TracesConfiguration` internal class to align with `MetricsConfiguration` and `LogsConfiguration`.
+
+### Features
+
 ## [1.1.0-beta.14] - 2026-08-23
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [1.1.0-beta.15-wip]
 
 ### Breaking Changes
-- `OTel.initialize` now always reads `OTEL_RESOURCE_ATTRIBUTES` directly, restoring spec-mandated environment variables for users who pass `detectPlatformResources: false`. The resource merging precedence has also been updated to strictly follow the OpenTelemetry specification (`Platform Resources` -> `OTEL_RESOURCE_ATTRIBUTES` -> `OTEL_SERVICE_NAME` -> `Code Attributes`).
 - **Spec compliance (#206):** `OTEL_RESOURCE_ATTRIBUTES` values are now parsed strictly as `String`s rather than falling back to `int` or `bool`.
 - Percent-decoding is now applied to `OTEL_RESOURCE_ATTRIBUTES` values.
 - Extracted trace exporter and processor configuration logic into `TracesConfiguration` internal class to align with `MetricsConfiguration` and `LogsConfiguration`.
--- a/lib/src/logs/export/logs_config.dart
+++ b/lib/src/logs/export/logs_config.dart
@@ -116,7 +116,7 @@ class LogsConfiguration {
 
     if (OTelLog.isDebug()) {
       OTelLog.debug('LogsConfiguration: Configured LoggerProvider with '
-          '${exporters.length} exporter(s) from OTEL_LOGS_EXPORTER');
+          '${createdExporters.length} exporter(s) from OTEL_LOGS_EXPORTER');
     }
 
     return logProvider;
--- a/lib/src/logs/export/logs_config.dart
+++ b/lib/src/logs/export/logs_config.dart
@@ -42,7 +42,14 @@ class LogsConfiguration {
     LogRecordExporter? logRecordExporter,
     LogRecordProcessor? logRecordProcessor,
     Resource? resource,
+    OtlpEnvironmentValues? otlpConfig,
+    List<String>? exporters,
+    BlrpEnvironmentValues? blrpConfig,
   }) {
+    otlpConfig ??= OTelEnv.getOtlpConfig(signal: 'logs');
+    exporters ??= OTelEnv.getExporters(signal: 'logs') ?? ['otlp'];
+    blrpConfig ??= OTelEnv.getBlrpConfig();
+
     // Get the logger provider
     final logProvider = OTel.loggerProvider();
 
@@ -59,16 +66,14 @@ class LogsConfiguration {
 
     // Explicitly provided exporter wins; otherwise read the env selection.
     if (logRecordExporter != null) {
-      logProvider.addLogRecordProcessor(_createProcessor(logRecordExporter));
+      logProvider.addLogRecordProcessor(
+          _createProcessor(logRecordExporter, blrpConfig));
       return logProvider;
     }
 
-    // Spec "Exporter Selection": OTEL_LOGS_EXPORTER, default otlp; the
-    // comma-separated list form is supported. Known: otlp, console, none.
     // Multiple exporters install one processor per exporter.
-    final requested = OTelEnv.getExporters(signal: 'logs') ?? ['otlp'];
-    if (requested.contains('none')) {
-      if (requested.length > 1 && OTelLog.isWarn()) {
+    if (exporters.contains('none')) {
+      if (exporters.length > 1 && OTelLog.isWarn()) {
         OTelLog.warn("OTEL_LOGS_EXPORTER contains 'none' alongside other "
             'values; installing no processor.');
       } else if (OTelLog.isDebug()) {
@@ -78,8 +83,8 @@ class LogsConfiguration {
       return logProvider;
     }
 
-    final exporters = <LogRecordExporter>[];
-    for (final name in requested) {
+    final createdExporters = <LogRecordExporter>[];
+    for (final name in exporters) {
       if (name == 'logging') {
         if (OTelLog.isWarn()) {
           OTelLog.warn("OTEL_LOGS_EXPORTER value 'logging' is deprecated "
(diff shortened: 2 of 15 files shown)


lib/src/otel.dart
michaelbushe, line 1: **Blocking — `OTel.initialize()` now throws on a malformed `OTEL_RESOURCE_ATTRIBUTES`.**

Calling `EnvVarResourceDetector().detect()` directly removes the per-detector `try`/`catch` that `CompositeResourceDetector.detect()` provided, so a value the percent-decoder can't read escapes initialization. Verified on both branches with `OTEL_RESOURCE_ATTRIBUTES='discount=50%'` and `detectPlatformResources: false`: main initializes fine (resource carries `service.name`, `service.version`, `discount`), this branch throws `ArgumentError: Truncated URI` from `Uri.decodeComponent`.

The blast radius is bigger than the exception. `OTelFactory.otelFactory` is assigned *before* resource detection, so the caller can't even retry — the second `initialize()` reports "can only be called once" while `defaultResource` is null and no span, metric or log processors were installed. One typo in an env var becomes an unrecoverable startup failure.

I've pushed failing tests in `46a7031` (`test/unit/environment/env_resource_attributes_resilience_test.dart`): three malformed values an operator could plausibly type, plus a well-formed control. 3 fail here, all 4 pass on main, so it's a regression rather than a pre-existing gap. Wrapping the `detect()` call in a `try`/`catch` (or routing it back through `CompositeResourceDetector`) makes them green.

Two more things in this file, non-blocking:

- **The env reads now happen before `sdkDisabled` is computed.** The comment this PR deletes said honoring `OTEL_SDK_DISABLED` early "keeps signal-specific configs (`OTEL_*_EXPORTER`) from being ever consulted," but `sdkDisabled` is now derived at line ~314 while lines ~171-178 already parsed OTLP config and exporter selection for all three signals. With `OTEL_SDK_DISABLED=true`, or `enableMetrics: false`, the SDK still parses those variables and emits warnings for malformed values of signals it will never configure.
- **`bspConfig` is read inline at its call site** rather than in the hoisted "Apply environment variables exactly once" block that `blrpConfig` and the rest follow — which undercuts this PR's own invariant and leaves the next person unsure where env reads belong.

lib/src/resource/resource_detector.dart
michaelbushe, line 1: **Blocking (public API) — `PlatformResourceDetector.create()` no longer includes `EnvVarResourceDetector`.**

`detectors = <ResourceDetector>[EnvVarResourceDetector()]` became `<ResourceDetector>[]`, and the compensating call was added only inside `OTel.initialize`. But `PlatformResourceDetector` is exported from the package barrel and is the documented way for an app or plugin to build its own resource. Any external caller doing `await PlatformResourceDetector.create().detect()` now silently loses `OTEL_RESOURCE_ATTRIBUTES` — no error, no warning.

Two existing tests named this contract and now assert nothing meaningful: `test/unit/sdk/resource_detector_test.dart:423` ("create() result includes EnvVarResourceDetector output") only checks `isNotNull`, and `test/web/web_compile_smoke_test.dart:64` still carries a comment saying the composite includes the env-var detector.

**Also blocking, now that this is the only parser: `_parseResourceAttributes` splits on every `=` and drops any pair whose value contains one.** Verified: `OTEL_RESOURCE_ATTRIBUTES='my.url=http://x/?a=b'` yields no `my.url` attribute at all, because `part.split('=')` returns three elements and the guard is `if (keyValue.length != 2) continue;`. The doc comment directly above says "Split on first equals sign", so the code and its own documentation disagree.

This was survivable while `getResourceAttributes()` also existed; this PR makes it the sole route by which the variable reaches the resource. `indexOf('=')` + `substring` is the idiom already used in `OTelEnv._parseHeaders` and `getServiceConfig` — and it's the same fix #199/#264 just landed for baggage.

Related, worth resolving in the same pass: after this PR the SDK has **three divergent parsers** for `OTEL_RESOURCE_ATTRIBUTES` — this one (splits on every `=`, percent-decodes, handles escaped commas, all Strings), `OTelEnv.getResourceAttributes` (now dead production code — no callers under `lib/`), and `getServiceConfig` (splits on first `=`, no decoding). Because `getServiceConfig`'s result merges *after* the detector's, they disagree on the same input: `service.name=my%20svc` ends up un-decoded in the final resource even though the detector decoded it correctly.

lib/src/trace/export/traces_config.dart
michaelbushe, line 1: Extracting this class is the right move and the main reason I want this PR to land — a few things before it does.

**It reproduces the bug #269 is removing.** `bool secure = false` passed to `resolveOtlpSecure` as `fallback:` is exactly the shape #253 filed against the logs and metrics configs: because `fallback` is the lowest-precedence input and every default endpoint carries a scheme, the parameter can never decide anything. #269 changes both siblings to `bool? secure` + `explicitSecure:`; the new traces config should match, or it becomes the one signal where the programmatic parameter is still a no-op.

**It ignores `otlpConfig.endpoint`.** `resolvedEndpoint = endpoint ?? (protocol == 'grpc' ? ... : ...)` never consults the config record it requires as a parameter, while `metric_config.dart` and `logs_config.dart` both do `otlpConfig.endpoint ?? endpoint ?? default`. A direct caller passing `otlpConfig: OTelEnv.getOtlpConfig(signal: 'traces')` with `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` set still exports to localhost. It only works from `OTel.initialize` because that pre-folds the env endpoint into `endpoint` — which makes the required parameter half-honored and the class fragile for exactly the reuse it was created to enable.

**It isn't exported from the barrel.** `logs_config.dart` and `metric_config.dart` are both public; this one is reachable only through a private relative import in `otel.dart`. So after a PR whose goal is consistency, traces becomes the only signal a user *cannot* configure manually. Worth exporting it, and considering `lib/src/trace/export/traces_config.dart` for symmetry with the siblings' placement and naming.

**Sampler assignment should be `??=`.** `OTel.tracerProvider()` already applies `sampler ??= _defaultSampler`, and line ~834 then unconditionally reassigns a second, independently constructed `ParentBasedSampler(const AlwaysOnSampler())`. The two expressions are identical today so nothing breaks, but the default now lives in two places that must stay in sync, and a caller who set `tracerProvider.sampler` before calling `configureTracerProvider(sampler: null)` has it clobbered.

**Minor:** `protocol`, `resolvedEndpoint`, `resolvedSecure` and the `buildOtlpExporter` closure are all computed before the `if (spanProcessor == null)` guard, so they're dead work whenever a caller supplies a processor (the common path in tests and for `SimpleSpanProcessor` users) or sets `OTEL_TRACES_EXPORTER=none`.

CHANGELOG.md
michaelbushe, line 1: The Breaking Changes entry describes a bug that didn't exist, and omits the three that do.

It says `detectPlatformResources: false` "completely bypassed the `EnvVarResourceDetector`, leading to spec-mandated environment variables being ignored." Verified against main: `OTel.initialize` folded `OTelEnv.getResourceAttributes()` in unconditionally, and a probe with `detectPlatformResources: false` shows the attribute present in the resulting resource. The precedence claim is also overstated — for non-service keys the relative order (platform < env < programmatic) is unchanged.

The three real breaking changes are undocumented:

1. **Resource attribute values change type.** `OTEL_RESOURCE_ATTRIBUTES='my.count=5,my.flag=true'` produced an `int` and a `bool` before and produces two `String`s now (verified by probe). That changes the OTLP wire type of every numeric and boolean resource attribute, which breaks backend queries, dashboards and alerts that filter on them. Worth saying explicitly: **this direction is correct** — the spec requires these values to be treated as strings, and it's filed as #206 — so the change should stay and simply be documented as breaking.
2. **Percent-decoding is now applied** to those values (`k=a%2Cb` → `a,b`), also correct and also undocumented.
3. **`PlatformResourceDetector.create()` no longer includes the env detector** — a public-API behavior change for anyone composing detectors themselves.

There's also a `### Features` heading with an empty body.

lib/src/trace/export/batch_span_processor.dart
michaelbushe, line 1: The validation logic survives the `fromEnvironment()` → `fromBspEnvironmentValues()` split intact — I checked — but two things got lost around it, and the logs twin in this same PR did both correctly, which is the useful comparison.

**The `OTEL_BSP_*` documentation table moved onto the factory that reads no environment.** It now sits above `fromBspEnvironmentValues(BspEnvironmentValues env)`, while `fromEnvironment()` — the one that actually consults those variables — has no doc comment at all. `batch_log_record_processor.dart` kept its table on `fromEnvironment`.

**Three spec-justification comments were deleted**: "Spec type: Duration. Zero is valid ('export as fast as possible')", "Spec type: Timeout. Zero means 'no limit' — substitute a very large duration", and "Spec rule: maxExportBatchSize must be less than or equal to maxQueueSize". The BLRP twin still has its copies. Without them, `if (timeout.inMilliseconds == 0) exportTimeout = noLimit;` reads as an arbitrary special case rather than the `Timeout`-vs-`Duration` distinction in `configuration/common.md` — which is a genuinely subtle rule, and one you got right in #72 by reading that section. Worth keeping the reasoning next to the code that depends on it.

lib/src/logs/export/logs_config.dart
michaelbushe, line 1: Small one from the rename: the debug log now reports the count of *requested* exporter names instead of the number actually created. The local list `exporters` became `createdExporters`, and a new parameter took the old name, but the log line still reads `${exporters.length}`. With `OTEL_LOGS_EXPORTER='otlp,logging,bogus'` one processor is installed and the log claims three. Should be `createdExporters.length`.

michaelbushe (changes requested):
This is the right refactor and I want it in — reading the environment once, and giving traces a `TracesConfiguration` to match metrics and logs, is exactly what #255 asked for. Running the `EnvVarResourceDetector` unconditionally is a genuine improvement too. Requesting changes on one blocker plus a few things that should not ship silently.

**The blocker: `OTel.initialize()` now throws on a malformed `OTEL_RESOURCE_ATTRIBUTES`.** Calling `EnvVarResourceDetector().detect()` directly loses the per-detector `try`/`catch` that `CompositeResourceDetector` provided. Verified on both branches with `discount=50%`: main initializes fine, this branch throws `ArgumentError: Truncated URI`. And because `OTelFactory` is installed before resource detection, the caller can't retry — the second `initialize()` says "can only be called once" while no processors were ever installed. I pushed failing tests to your branch (`46a7031`); 3 fail here and all 4 pass on main, so it's a regression, and a `try`/`catch` around the call turns them green.

**Then, in rough priority order:**

1. `PlatformResourceDetector.create()` is public API and no longer includes the env detector, so external callers composing their own resources silently lose `OTEL_RESOURCE_ATTRIBUTES` (two existing tests that named this contract now assert only `isNotNull`).
2. `_parseResourceAttributes` splits on every `=` and drops values containing one — verified: `my.url=http://x/?a=b` disappears entirely. It's now the *only* parser for that variable, and its own doc comment says "split on first equals sign". Same fix as #199/#264: `indexOf('=')` + `substring`.
3. The CHANGELOG describes a bug that didn't exist and omits three real breaking changes — including the value type change from `int`/`bool` to `String`, which is **correct** per spec (#206) and just needs documenting.
4. `TracesConfiguration` reproduces the #253 `bool secure = false` + `fallback:` bug that #269 is removing from its siblings, ignores its own `otlpConfig.endpoint`, and isn't exported from the barrel — so after a consistency PR, traces is the only signal a user can't configure manually.

Details are in the file comments. Sequencing note: #269 touches `otel.dart`, `logs_config.dart` and `metric_config.dart` too and fixes a TLS-disabling bug, so I'd like that one to land first and have this rebase onto it — at which point picking up `bool? secure` in `TracesConfiguration` is a two-line change. Issue #270 (named providers export nothing) then builds directly on this PR's `TracesConfiguration`, so this work unblocks it.

Review notes: maintainer-reviewed, AI-assisted analysis; the regressions above were reproduced by running both branches, not inferred from reading.

harshitt13:
I've wired the `exemplarFilter` through configuration down to `MetricTransformer` in this PR. 

As for #154, I agree—let's handle the actual application of the filter in a separate follow-up PR before the next release so this one stays focused.

michaelbushe:
Thank you @harshitt13! Nice work.

PR #271, opened 2026-08-25
fix(propagation): correct W3C traceparent parsing

### LINKED ISSUE #192: [BUG] Propagation: traceparent with a future version is rejected instead of parsed
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Propagators API, W3C Trace Context Requirements
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/context/api-propagators/#w3c-trace-context-requirements
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/context/api-propagators.md#w3c-trace-context-requirements

> A W3C Trace Context propagator MUST parse and validate the `traceparent` and `tracestate` HTTP headers as specified in [W3C Trace Context Level 2](https://www.w3.org/TR/trace-context-2/).

W3C Trace Context Level 2, section "Versioning of traceparent", says that an implementation that receives a higher version SHOULD try to parse it with the highest version it knows. It reads the first four version-00 fields and allows extra trailing data after the trace-flags field, unless the version is `ff`.

## The problem

The propagator rejects every header that is not exactly version `00`. A comment in the code admits the deviation.

Two checks combine.

The length check requires exactly 55 characters. A future-version header carries extra fields after trace-flags, which W3C allows. That header fails here.

The version check then rejects any version other than `00`. A header such as `cc-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01` is dropped, although W3C says to parse it as version 00.

Trace continuity therefore breaks as soon as an upstream service emits a newer version.

## Offending code

`dartastic_opentelemetry/lib/src/trace/w3c_trace_context_propagator.dart:137` and `:162`

```dart
if (traceparent.length != _traceparentLength) { // exactly 55
  ...
  return null;
}
...
// Validate version (currently only 00 is supported)
if (version != _version) {
  ...
  // Per spec, we should still try to parse if version is unknown
  // but for now we'll reject it
  return null;
}
```

## How to correct it

1. Accept a version field that matches `^[0-9a-f]{2}$` and is not `ff`.
2. For version `00`, require exactly 55 characters.
3. For a higher version, require at least 55 characters, then require either the end of the string or a `-` at index 55.
4. Parse the first four fields with the version 00 rules.

## Notes

- Found by: Fable 5 (both runs), Grok 4.5, GPT-5.6 Sol
- Source finding files: `fable-ultra/context-03-future-traceparent-version-rejected.md`, `fable-medium/propagation-traceparent-future-version.md`, `grok/trace/017-w3c-rejects-unknown-versions.md`, `codex/context-w3c-future-traceparent-versions.md`
- Parent issue: #95

### LINKED ISSUE #193: [BUG] Propagation: traceparent accepts uppercase and signed hex in trace-id and span-id
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Propagators API, W3C Trace Context Requirements
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/context/api-propagators/#w3c-trace-context-requirements
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/context/api-propagators.md#w3c-trace-context-requirements

> A W3C Trace Context propagator MUST parse and validate the `traceparent` and `tracestate` HTTP headers as specified in [W3C Trace Context Level 2](https://www.w3.org/TR/trace-context-2/).

W3C Trace Context defines `trace-id` and `parent-id` as lowercase hex only. A header that holds a character outside that set is invalid, and a vendor must ignore it.

## The problem

`_parseTraceparent` checks field lengths only. It then hands the hex to the ID helpers, which call `int.tryParse` with radix 16 on each two-character slice.

`int.tryParse` accepts uppercase hex. It also accepts a sign character. `'AB'` gives 171. `'+f'` gives 15. `'-1'` gives -1, which wraps to 255 when the code stores it in a `Uint8List`. Fable verified this behavior directly.

A header such as `00-4BF92F3577B34DA6A3CE929D0E0E4736-00F067AA0BA902B7-01` is therefore accepted. W3C calls it invalid. The extract rules then require the Context to stay unchanged.

## Offending code

`dartastic_opentelemetry/lib/src/trace/w3c_trace_context_propagator.dart:203`

```dart
final traceId = OTel.traceIdFrom(traceIdHex);
final spanId = OTel.spanIdFrom(spanIdHex);
```

`dartastic_opentelemetry_api/lib/src/api/id/id_generator.dart:62`

```dart
final byte = int.tryParse(hexByte, radix: 16);
```

## How to correct it

Validate each field against the grammar in `_parseTraceparent`, before any conversion. Use `^[0-9a-f]{32}$` for the trace ID, `^[0-9a-f]{16}$` for the span ID, and `^[0-9a-f]{2}$` for the version and the trace flags. Return null on any mismatch.

Rejecting a character outside `[0-9a-f]` inside `IdGenerator.hexToBytes` also works.

NOTE: The trace-flags issue in this same section shares this parse path. Fix both together.

## Notes

- Source finding files: `fable-ultra/context-01-traceparent-hex-validation.md`, `fable-medium/propagation-traceparent-field-validation.md`
- Parent issue: #95

### LINKED ISSUE #194: [BUG] Propagation: invalid trace-flags are coerced to 0x00 instead of rejecting the header
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Propagators API, W3C Trace Context Requirements
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/context/api-propagators/#w3c-trace-context-requirements
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/context/api-propagators.md#w3c-trace-context-requirements

> A W3C Trace Context propagator MUST parse and validate the `traceparent` and `tracestate` HTTP headers as specified in [W3C Trace Context Level 2](https://www.w3.org/TR/trace-context-2/).

W3C Trace Context defines `trace-flags` as two lowercase hex digits. When any field holds a character outside the allowed set, the whole header is invalid and a vendor must not use it. The vendor restarts the trace instead.

## The problem

`_parseTraceparent` checks the length of the trace-flags field only. It then calls `TraceFlags.fromString`.

That factory swallows a parse failure and substitutes "no flags".

A header such as `00-<valid-trace-id>-<valid-span-id>-zz` is therefore accepted as a valid remote span context with flags `0x00`. W3C requires the whole header to be rejected.

Two things go wrong. The validation requirement is not met, and the sampled bit is silently dropped. A restarted trace would have decided that bit for itself.

## Offending code

`dartastic_opentelemetry/lib/src/trace/w3c_trace_context_propagator.dart:205`

```dart
final traceFlags = TraceFlags.fromString(traceFlagsHex);
```

`dartastic_opentelemetry_api/lib/src/api/trace/trace_flags.dart:48`

```dart
factory TraceFlags.fromString(String hex) {
  final flags = int.tryParse(hex, radix: 16) ?? NONE_FLAG;
  return TraceFlagsCreate.create(flags);
}
```

## How to correct it

Check `traceFlagsHex` against `^[0-9a-f]{2}$` inside `_parseTraceparent`. Return null on a mismatch. Do not rely on the lenient factory.

NOTE: Uppercase flags pass today for the same reason as the hex validation issue in this same section. One regular expression fixes both.

## Notes

- Source finding files: `fable-ultra/context-02-trace-flags-invalid-coerced.md`, `fable-medium/propagation-traceparent-field-validation.md`
- Parent issue: #95

robert-northmind:
## Summary

Fixes three W3C Trace Context Level 2 compliance issues:

- **#192:** Parse higher `traceparent` versions using the version `00` field
  layout, while retaining version `00`'s strict grammar.
- **#193:** Reject version, trace ID, and parent ID fields that are not
  fixed-width lowercase hexadecimal.
- **#194:** Reject malformed trace flags instead of coercing them to unsampled.

## Changes

- Reworked `traceparent` parsing around fixed field offsets.
- Added focused W3C Level 2 contract coverage, including higher-version
  downgrade behavior.
- Updated existing propagator and debug-log tests.
- Added the fixes to the current changelog section.

## Verification

- `dart format` clean
- `dart analyze` clean
- Relevant contract and infrastructure tests: 84 passed
- Full test and coverage suite passed via the pre-push checks
- Line coverage: 93.5914% (required: 93.4%)
- Bugbot: no bugs

Closes #192
Closes #193
Closes #194

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.1.0-beta.15-wip]
 
+### Fixed
+
+- **`traceparent` parsing now follows W3C Trace Context Level 2** (#192, #193,
+  #194). Higher versions from `01` through `fe` are parsed using the version
+  `00` field layout and may carry trailing fields, while version `00` retains
+  its strict grammar. Version, trace ID, parent ID, and trace flags now require
+  fixed-width lowercase hexadecimal syntax, preventing malformed identifiers
+  from being accepted and invalid flags from being coerced to unsampled.
+
 ## [1.1.0-beta.14] - 2026-08-23
 
 ### Changed
--- a/lib/src/trace/w3c_trace_context_propagator.dart
+++ b/lib/src/trace/w3c_trace_context_propagator.dart
@@ -29,11 +29,34 @@ class W3CTraceContextPropagator
   /// The standard header name for W3C trace state
   static const _tracestateHeader = 'tracestate';
 
-  /// The current version of the W3C Trace Context specification
+  /// The highest traceparent version this propagator understands, and the
+  /// version it emits. Per the spec's versioning rules, higher versions are
+  /// parsed using this version's field layout.
   static const _version = '00';
 
-  /// The length of a valid traceparent header value
+  /// The version reserved as invalid by the specification.
+  static const _forbiddenVersion = 'ff';
+
+  static const _versionLength = 2;
+  static const _traceIdLength = 32;
+  static const _spanIdLength = 16;
+  static const _traceFlagsLength = 2;
+
+  /// Field offsets, applied to every version. The spec assumes future versions
+  /// will be additive to version 00, and on that basis defines the parse of a
+  /// higher version as reading these same positions.
+  static const _traceIdOffset = _versionLength + 1;
+  static const _spanIdOffset = _traceIdOffset + _traceIdLength + 1;
+  static const _traceFlagsOffset = _spanIdOffset + _spanIdLength + 1;
+
+  /// The exact length of a version 00 traceparent, and the minimum length of
+  /// any traceparent.
   static const _traceparentLength = 55; // 00-{32}-{16}-{2}
+
+  /// The specification's grammar allows lowercase hex only.
+  static final _hexPairPattern = RegExp(r'^[0-9a-f]{2}$');
+  static final _traceIdPattern = RegExp('^[0-9a-f]{$_traceIdLength}\$');
+  static final _spanIdPattern = RegExp('^[0-9a-f]{$_spanIdLength}\$');
 
   @override
   Context extract(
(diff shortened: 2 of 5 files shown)


robert-northmind:
Hey @michaelbushe, I started working on fixing the #192 and while doing so, the #193 & #194 seemed very related and didn't add to much extra code to this PR. So I fixed them all in one go.
But as I opened this PR, I saw you comment in the 193 & 194 about fixing them in the API layer first.
I did some searching around and I think the fix belongs in the `trace_context_propegator`. Seems to be how other OTel libs are doing it. But I might be missing something.

What was the thought of doing the API fix first? 
From the issues in the API repo, it also seems the preferred way is to fix the issue in the `_parseTraceparent` in the SDK.

michaelbushe:
The IdGenerator.hexToBytes is in the API. We should fix the API first.  In my opinion, we should fix *all* the API issues, then move onto the SDK.

Issue #219, opened 2026-08-15
[BUG] OTLP: signal-specific endpoint URLs are modified instead of used as-is


yuzurihaaa:
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — OTLP Exporter, Endpoint URLs for OTLP/HTTP
**Requirement level:** MUST / MUST NOT
**URL:** https://opentelemetry.io/docs/specs/otel/protocol/exporter/#endpoint-urls-for-otlphttp
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/protocol/exporter.md#endpoint-urls-for-otlphttp

> For the per-signal variables (`OTEL_EXPORTER_OTLP_<signal>_ENDPOINT`), the URL MUST be used as-is without any modification. The only exception is that if an URL contains no path part, the root path `/` MUST be used (see Example 2).

> An SDK MUST NOT modify the URL in ways other than specified above.

## The problem

All three OTLP/HTTP exporters append the signal path to whatever endpoint they receive. None of them knows where that endpoint came from.

A per-signal variable must be used as it is. The base variable takes the appended path. `OTelEnv.getOtlpConfig` collapses both into one value and discards which variable supplied it.

Three concrete failures follow.

1. `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://collector:4318` sends to `.../v1/traces`. The spec requires the root path. Example 2 in the spec shows this exact case.
2. `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://collector:4318/custom/path` becomes `.../custom/path/v1/traces`. The URL must not change.
3. The suffix check skips the append when a base endpoint legitimately ends in `/v1/traces` as a prefix path.

## Offending code

`dartastic_opentelemetry/lib/src/trace/export/otlp/http/otlp_http_span_exporter.dart:79-90`

```dart
String _getEndpointUrl() {
  // Ensure the endpoint ends with /v1/traces
  var endpoint = _config.endpoint;
  if (!endpoint.endsWith('/v1/traces')) {
    if (endpoint.endsWith('/')) {
      endpoint = endpoint.substring(0, endpoint.length - 1);
    }
    endpoint = '$endpoint/v1/traces';
  }
  return endpoint;
}
```

[...]

PR #273, opened 2026-08-25
test: add regression test for OTEL_EXPORTER_OTLP_METRICS_ENDPOINT (#229)

### LINKED ISSUE #229: [BUG] OTLP: OTEL_EXPORTER_OTLP_METRICS_ENDPOINT is read but never applied
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — OTLP Exporter, Configuration Options
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/protocol/exporter/#configuration-options
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/protocol/exporter.md#configuration-options

> The following configuration options MUST be available to configure the OTLP exporter. Each configuration option MUST be overridable by a signal specific option.

> Env vars: `OTEL_EXPORTER_OTLP_ENDPOINT` `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT`

## The problem

`MetricsConfiguration._createExporter` reads the metrics signal configuration. It never uses the endpoint from it.

The exporter always receives the endpoint parameter handed down from `OTel.initialize()`, which came from the traces and global configuration.

Setting `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` therefore does nothing. Metrics go to the traces endpoint.

The logs path does apply its own value. That difference marks this as an omission, not a decision.

## Offending code

`dartastic_opentelemetry/lib/src/metrics/export/metric_config.dart:122-192`

```dart
final otlpConfig = OTelEnv.getOtlpConfig(signal: 'metrics');
final protocol = otlpConfig['protocol'] as String? ?? 'http/protobuf';
...
return OtlpHttpMetricExporter(
  OtlpHttpMetricExporterConfig(
    endpoint: endpoint,   // <- parameter from otel.dart, not otlpConfig['endpoint']
    ...
```

The correct pattern is at `logs/export/logs_config.dart:129`.

## How to correct it

Mirror the logs path:

```dart
final effectiveEndpoint = otlpConfig['endpoint'] as String? ?? endpoint;
```

Pass that value to both metric exporter configs. Pass it to `resolveOtlpSecure` as well, which reads the wrong endpoint today at `metric_config.dart:146-150`.

## Notes

- Source finding files: `fable-medium/otlp-metrics-endpoint-env-ignored.md`, `fable-ultra/config-10-metrics-endpoint-env-ignored.md`
- Parent issue: #95

abidiahmedcom:
Closes #229

The fix for this issue is already in `metric_config.dart:149`:

```dart
final effectiveEndpoint = otlpConfig.endpoint ?? endpoint ?? ...
```

`otlpConfig.endpoint` reads from `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` via
`OTelEnv.getOtlpConfig(signal: 'metrics')`. The env var is applied correctly
for both HTTP and gRPC metric exporters.

This PR adds two regression tests to lock in that behavior and prevent
future regressions:

- `exporter_selection_inprocess_test.dart`: 2 new tests in a new
  `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` group
- `CHANGELOG.md`: regression test entry for #229

Local verification:
- `dart analyze`: 0 issues
- `dart format`: no changes
- `dart test exporter_selection_inprocess_test.dart`: 18/18 pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Previously `secure` overrode the gRPC endpoint scheme; now the scheme
   (`http:`, `https:`) takes precedence and `secure` applies only to
   scheme-less endpoints.
+- Regression test verifying `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` is applied
+  to both HTTP and gRPC metric exporters. ([#229](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/issues/229))
 
 ## [1.1.0-beta.14] - 2026-08-23
 
--- a/test/unit/environment/exporter_selection_inprocess_test.dart
+++ b/test/unit/environment/exporter_selection_inprocess_test.dart
@@ -45,6 +45,33 @@ void main() {
     EnvironmentService.testOverrides = null;
   });
 
+  group('OTEL_EXPORTER_OTLP_METRICS_ENDPOINT', () {
+    test('signal-specific metrics endpoint flows to the http exporter',
+        () async {
+      OTelLog.currentLevel = LogLevel.debug;
+      await initWith({
+        'OTEL_EXPORTER_OTLP_METRICS_ENDPOINT': 'http://custom-metrics:9999',
+      }, metrics: true);
+      final reader = OTel.meterProvider().metricReaders.single
+          as PeriodicExportingMetricReader;
+      expect(reader.exporter, isA<OtlpHttpMetricExporter>());
+      expect(warnings.join('\n'), contains('http://custom-metrics:9999'));
+    });
+
+    test('signal-specific metrics endpoint flows to the grpc exporter',
+        () async {
+      OTelLog.currentLevel = LogLevel.debug;
+      await initWith({
+        'OTEL_EXPORTER_OTLP_METRICS_ENDPOINT': 'http://custom-metrics:4317',
+        'OTEL_EXPORTER_OTLP_PROTOCOL': 'grpc',
+      }, metrics: true);
+      final reader = OTel.meterProvider().metricReaders.single
+          as PeriodicExportingMetricReader;
+      expect(reader.exporter, isA<OtlpGrpcMetricExporter>());
+      expect(warnings.join('\n'), contains('http://custom-metrics:4317'));
+    });
+  });
+
   group('OTEL_METRICS_EXPORTER selection', () {
     test('none installs no reader', () async {
       await initWith({'OTEL_METRICS_EXPORTER': 'none'}, metrics: true);


michaelbushe (changes requested):
Requesting changes on one point. First, credit where it is due: I checked the thing that actually decides whether a regression test is worth having, and this one passes it.

Reverting the fix on current `main` — removing `otlpConfig.endpoint ??` from `metric_config.dart:159` so the signal-specific variable is ignored again — produces:

```
fix present:   18 tests, all pass
fix reverted:  16 pass, 2 fail   <- exactly the two new tests, nothing else
```

Correctly scoped, and it fails for the right reason. The description is also honest about what the change is: it states up front that the fix already exists and that this is coverage only. That is the right way to submit this.

## Change requested: the assertion is coupled to a debug log line, not to the behavior

```dart
OTelLog.currentLevel = LogLevel.debug;          // harness default is warn
...
expect(warnings.join('\n'), contains('http://custom-metrics:9999'));
```

The log level is escalated to `debug` precisely because the endpoint only ever appears in an `OTelLog.debug` message. That makes the log wording part of the contract under test. Verified: deleting *only* the

```dart
OTelLog.debug('MetricsConfiguration: Creating OtlpHttpMetricExporter for $effectiveEndpoint');
```

statement — behavior completely unchanged, the endpoint still correctly applied — fails the HTTP test. A future cleanup of that log line breaks a test that is supposed to be guarding endpoint resolution, and the failure will point at the wrong thing.

Assert the resolved endpoint directly instead:

- **gRPC** works today. `OtlpGrpcMetricExporter.config` is already `@visibleForTesting` (exposed for #253), so:
  ```dart
  final exporter = reader.exporter as OtlpGrpcMetricExporter;
  expect(exporter.config.endpoint, 'http://custom-metrics:4317');
  ```
- **HTTP** needs the mirror-image exposure first. `OtlpHttpMetricExporter` keeps `_config` private; add the same `@visibleForTesting final OtlpHttpMetricExporterConfig config;` field that the gRPC exporter has, then assert `config.endpoint`.

With that, `OTelLog.currentLevel = LogLevel.debug` can come out of both tests and the assertions stop depending on log text.

## Minor

- The description cites `metric_config.dart:149`; it is line 159 on `main`. Same root cause as the merge conflict — the branch is based on an older `main`. Please rebase; the conflict is `CHANGELOG.md` only.
- The CHANGELOG entry uses `([#229](https://github.com/.../issues/229))`. This file uses bare `(#229)` in 44 places and a markdown link in 3. Match the majority.
- Worth correcting in the description for the record: the fix did not come from the recent metrics configuration work. `git log -S effectiveEndpoint -- lib/src/metrics/export/metric_config.dart` returns exactly one commit — `0e86f6c`, "Fix OTLP spec-compliance issues (#213, #220, #228) (#238)", 2026-08-21. #229 was fixed incidentally there and the issue was never closed, which is why this coverage was missing.

Nothing here is structural. Make the assertion direct and rebase, and this is good to merge.

michaelbushe:
Thanks for looking at this. I'm going to close it. #229 was already fixed on main in #72, so this pins existing behavior rather than guarding a change, and the assertions match against debug log text rather than exporter behavior — which couples them to a log string we've already rewritten once. We're adding end-to-end tests that assert on the request path a collector actually receives; those cover this case and also catch #219, where a per-signal endpoint is still modified instead of used as-is.

abidiahmedcom:
@michaelbushe Thanks for the thorough review. Here is what changed:

**Direct endpoint assertions**: Removed the log-level setup and replaced expect(warnings.join('\n'), ...) with direct config assertions. Added @visibleForTesting to OtlpHttpMetricExporter.config (mirroring the gRPC exporter) so the HTTP test asserts exporter.config.endpoint directly.

**Rebased** on current main. The CHANGELOG conflict is resolved.

**CHANGELOG format**: Fixed to bare (#229) to match the 44 existing entries.

One note on the attribution: you are right that the fix came from #238. I updated the PR description to make that clear.

PR #276, opened 2026-08-26
fix: service.name defaults to unknown_service per OTel spec (#204)

### LINKED ISSUE #204: [BUG] Resource: service.name falls back to a package name instead of unknown_service
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Resource SDK, SDK-provided resource attributes
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/resource/sdk/#sdk-provided-resource-attributes
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/resource/sdk.md#sdk-provided-resource-attributes

> The SDK MUST provide access to a Resource with at least the attributes listed at [Semantic Attributes with SDK-provided Default Value](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/README.md#semantic-attributes-with-sdk-provided-default-value). This resource MUST be associated with a `TracerProvider`, `MeterProvider`, or `LoggerProvider` if another resource was not explicitly specified.

The referenced semantic conventions require `service.name` with the SDK-provided default `unknown_service:<process.executable.name>`. The default is `unknown_service` when the executable name is not available.

## The problem

When no service name arrives from the caller, from `OTEL_SERVICE_NAME`, or from `OTEL_RESOURCE_ATTRIBUTES`, the SDK uses a vendor package name.

A search for `unknown_service` across the SDK returns no result.

The SDK also invents a `service.version` of `1.0.0`. That attribute is not one of the SDK-provided defaults, and the invented value misstates the version of the user's service.

## Offending code

`dartastic_opentelemetry/lib/src/otel.dart:70`

```dart
/// Default service name used if none is provided.
static const defaultServiceName = '@dart/dartastic_opentelemetry';
```

The fallback runs at `otel.dart:177`. The invented version is at `otel.dart:178`. Both reach the default resource at `otel.dart:278-284`.

## How to correct it

1. Default `service.name` to `unknown_service:<process.executable.name>` on a native platform. Take the base name of `Platform.executable`.
2. Default it to `unknown_service` where the executable name is not available, such as the web.
3. Do not set a `service.version` when the caller gives none.

## Notes

- Source finding files: `fable-ultra/config-06-service-name-fallback.md`, `fable-medium/resource-default-sdk-attributes.md`
- Parent issue: #95

abidiahmedcom:
Closes #204

The OTel specification requires the default `service.name` to be
`unknown_service:<process.executable.name>`. The SDK was using a
vendor-specific name (`@dart/dartastic_opentelemetry`) and inventing
a `service.version` of `1.0.0` when none was provided.

This fix:

1. Changes `defaultServiceName` from a const to a computed getter that
   returns `unknown_service:<executable>` on native platforms and
   `unknown_service` on web (where `dart:io` is unavailable).

2. Removes the `serviceVersion ??= '1.0.0'` default. The version is now
   only set when the caller explicitly provides one (via argument,
   `OTEL_SERVICE_VERSION`, or `OTEL_RESOURCE_ATTRIBUTES`).

3. Uses conditional imports for `Platform.executable` access:
   - `executable_name.dart` (barrel with platform conditional export)
   - `executable_name_io.dart` (native: reads `Platform.executable`)
   - `executable_name_web.dart` (web: returns empty string)

Changes:
- `otel.dart`: `defaultServiceName` is now a computed getter, `serviceVersion`
  stays nullable, resource attributes conditionally include `service.version`
- `environment/executable_name*.dart`: new platform-conditional files
- `otel_service_precedence_test.dart`: updated to expect null `service.version`
  when not provided
- `CHANGELOG.md`: entry under Fixed

Local verification:
- `dart analyze`: 0 issues
- `dart format`: no changes
- 1000+ tests pass (2 pre-existing metric failures unrelated to this change)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `false`**, and a malformed value is dropped with a warning instead of
   failing `OTel.initialize`.
 
+- **`service.name` defaults to spec-compliant `unknown_service`** (#204):
+  Previously defaulted to `@dart/dartastic_opentelemetry`. Now defaults to
+  `unknown_service:<process.executable.name>` on native platforms and
+  `unknown_service` on web, as required by the OTel specification.
+  `service.version` is no longer invented when not provided.
+
 ## [1.1.0-beta.14] - 2026-08-23
 
 ### Changed
--- a/lib/src/environment/executable_name.dart
+++ b/lib/src/environment/executable_name.dart
@@ -1,0 +1,5 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+export 'executable_name_io.dart'
+    if (dart.library.js_interop) 'executable_name_web.dart';
--- a/lib/src/environment/executable_name_io.dart
+++ b/lib/src/environment/executable_name_io.dart
@@ -1,0 +1,12 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:io' as io;
+
+/// Returns the base name of the current process executable.
+String get processExecutableName {
+  final path = io.Platform.executable;
+  // Extract just the file name from the full path.
+  final lastSlash = path.lastIndexOf(io.Platform.pathSeparator);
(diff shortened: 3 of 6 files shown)


abidiahmedcom:
Closing per maintainer guidance.

Issue #270, opened 2026-08-24
[BUG] Named providers (addTracerProvider/addMeterProvider/addLoggerProvider) export nothing — config only ever targets the default provider


michaelbushe:
Named providers created with `OTel.addTracerProvider`, `OTel.addMeterProvider` or `OTel.addLoggerProvider` are returned with **no processors, readers or exporters attached**. Telemetry recorded through them is created, ended, and silently discarded — no export, no warning.

## Reproduction

```dart
await OTel.initialize(serviceName: 'demo');           // default provider works
final audit = OTel.addLoggerProvider('audit');        // no processors attached
audit.getLogger('audit').info('this is never exported');
```

The same holds for `addTracerProvider` (spans) and `addMeterProvider` (metrics).

## Cause

All exporter/processor wiring targets the **default** provider only:

| Signal  | Where the wiring happens                                            | Targets                    |
|---------|---------------------------------------------------------------------|----------------------------|
| Traces  | `OTel.initialize` → `OTel.tracerProvider().addSpanProcessor(...)`    | default only               |
| Metrics | `MetricsConfiguration.configureMeterProvider` → `OTel.meterProvider()` | default only             |
| Logs    | `LogsConfiguration.configureLoggerProvider` → `OTel.loggerProvider()`  | default only             |

The configuration functions accept `endpoint`, `secure`, exporters and processors as parameters, but fetch the provider they configure instead of receiving it — so every parameter is implicitly scoped to the global default. The named factories (`addLoggerProvider` and friends) construct a provider, set its resource, and return.

Notably, all three factories already accept an `endpoint` parameter that currently has no effect on export, which suggests per-provider configuration was the original intent and was simply never wired.

## Not a regression

This is original behavior, not something recently introduced: the logs path dates to #24 (2026-04-05), the metrics path to the initial revision, and `addLoggerProvider` has never wired processors. Recent env-parsing work (#251, #66) touched these files but never the provider fetch.

## Fix

1. Give each configuration an optional target: `configureXProvider({XProvider? provider, ...})`, using `provider ?? OTel.xProvider()` so existing callers are unaffected.
2. Have `addTracerProvider` / `addMeterProvider` / `addLoggerProvider` call their configuration with the newly created provider, forwarding that provider's `endpoint`.
3. Add `bool? secure` to the three factories for parity with the programmatic `secure` parameter (#253 / #269).

[...]

Issue #153, opened 2026-08-15
[BUG] Metrics SDK: View stream configuration is missing required parameters


yuzurihaaa:
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Metrics SDK, Stream configuration
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/metrics/sdk/#stream-configuration
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/sdk.md#stream-configuration

> The SDK MUST accept the following stream configuration parameters:
>
> * `name`, `description`, `attribute_keys`, `aggregation`
> * `exemplar_reservoir`: ... If the user does not provide an `exemplar_reservoir` value, the `MeterProvider` MUST apply a [default exemplar reservoir](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#exemplar-defaults).
> * `aggregation_cardinality_limit`: A positive integer value defining the maximum number of data points allowed to be emitted in a collection cycle by a single instrument.

## The problem

`View` exposes `name`, `description`, `aggregationType`, and `attributeKeys` only.

Four parameters are missing:

1. No `exemplar_reservoir` parameter. The SDK has no reservoir concept at all.
2. No `aggregation_cardinality_limit` parameter. The SDK has no cardinality limit concept at all.
3. No way to configure an aggregation. `AggregationType.histogram` is an enum member with no fields. A View therefore cannot set bucket boundaries or `RecordMinMax`.
4. No exclude-list for attribute keys.

## Offending code

`dartastic_opentelemetry/lib/src/metrics/view.dart:31-71`. The parameterless enum is at `view.dart:15`.

## How to correct it

1. Replace the `AggregationType` enum with aggregation objects that carry configuration. One example is `ExplicitBucketHistogramAggregation(boundaries:, recordMinMax:)`.
2. Add an optional `exemplarReservoirFactory` field.
3. Add an optional `aggregationCardinalityLimit` field.
4. Add an `excludedAttributeKeys` list.
5. Fall back to the MetricReader default when a field is absent.

NOTE: Items 2 and 3 need the exemplar issue and the cardinality issue in this same section. Fix those first.

## Notes

- Spec version: **v1.60.0**. The audit ran against v1.59.0. The quoted text above was confirmed word for word in the v1.60.0 source on 2026-08-15.
[...]

Issue #151, opened 2026-08-15
[BUG] Metrics SDK: Views can be registered but are never applied


yuzurihaaa:
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Metrics SDK, View
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/metrics/sdk/#view
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/sdk.md#view

> The SDK MUST provide functionality for a user to create Views for a `MeterProvider`. This functionality MUST accept as inputs the [Instrument selection criteria](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#instrument-selection-criteria) and the resulting [stream configuration](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#stream-configuration).
>
> The SDK MUST provide the means to register Views with a `MeterProvider`.

From Stream configuration:

> The allow-list contains attribute keys that identify the attributes that MUST be kept, and all other attributes MUST be ignored.

## The problem

A user can create a `View` and register it. No other code ever reads it.

The `_views` list has two readers only. One is the `views` getter. The other clears the list at shutdown.

`Meter`, the instruments, and `collectAllMetrics` never call `View.matches`. They never read `View.name`, `View.description`, `View.aggregationType`, or `View.attributeKeys`.

Three results follow:

1. Renaming a stream, overriding a description, and filtering attribute keys do nothing.
2. `AggregationType.drop` drops nothing. The required Drop Aggregation therefore does not exist.
3. A user cannot select an aggregation for an instrument. Every instrument keeps its fixed storage.

The `View` class is dead configuration.

## Offending code

`dartastic_opentelemetry/lib/src/metrics/meter_provider.dart:158`

```dart
void addView(View view) {
  _views.add(view);
}
```

The collection path is at `meter_provider.dart:204`. `View.matches` sits at `view.dart:95` with no caller in production code.

## How to correct it

1. Match every registered View against an instrument when the code creates the instrument, or at the first collection.
2. Build one metric stream for each match. Apply the name, the description, the attribute key filter, and the aggregation.
3. Make Drop suppress the stream.
4. Use the default aggregation when no View matches.

NOTE: This is the head of the Views cluster. See the selection criteria issue and the stream configuration issue in this same section.

## Notes

[...]

PR #277, opened 2026-08-26
feat: Implement Metrics Exemplar support per OTel v1.60.0 (#154)

### LINKED ISSUE #154: [BUG] Metrics SDK: no ExemplarFilter or ExemplarReservoir, exemplars are never sampled

harshitt13:
Resolves #154 

This PR implements Metrics SDK Exemplar support according to the OpenTelemetry Specification v1.60.0. This allows the metrics SDK to collect specific exemplar data points along with aggregated metrics, establishing a correlation between metrics and trace data.

### Changes
- **Exemplar Filter Integration:** Added `ExemplarFilter` interfaces and implemented the standard filters (`AlwaysOnExemplarFilter`, `AlwaysOffExemplarFilter`, `TraceBasedExemplarFilter`). Plumbed `OTEL_METRICS_EXEMPLAR_FILTER` through `MetricsConfiguration`.
- **Exemplar Reservoirs:** Introduced `ExemplarReservoir` sampling strategies (`SimpleFixedSizeExemplarReservoir` for monotonic sums and gauges, and `AlignedHistogramBucketExemplarReservoir` for histograms).
- **Storage Refactoring:** Removed manual `addExemplar` injection and updated `SumStorage`, `GaugeStorage`, and `HistogramStorage` to automatically sample exemplars via their internal reservoirs during the `record()` cycle.
- **Instrument Context Capture:** Upgraded all metric instruments (Counters, Histograms, Gauges, and Observables) to automatically capture `Context.current` during `record()` and pass it to the storage layer for tracing correlations.
- **Testing:** Fixed and enhanced tests to thoroughly cover `Exemplar.fromMeasurement` mapping and ensuring reservoirs correctly retain valid trace metadata without memory leaks.

### DIFF
--- a/lib/src/metrics/data/exemplar.dart
+++ b/lib/src/metrics/data/exemplar.dart
@@ -43,7 +43,8 @@ class Exemplar {
 
   /// Creates an exemplar from a measurement.
   factory Exemplar.fromMeasurement({
-    required Measurement measurement,
+    required num value,
+    required Attributes measurementAttributes,
     required DateTime timestamp,
     required Attributes aggregationAttributes,
     SpanId? spanId,
@@ -51,7 +52,7 @@ class Exemplar {
   }) {
     // Determine which attributes are filtered out
     final filteredAttrs = _filterAttributes(
-      measurement.attributes ?? OTelFactory.otelFactory!.attributes(),
+      measurementAttributes,
       aggregationAttributes,
     );
 
@@ -59,7 +60,7 @@ class Exemplar {
       attributes: aggregationAttributes,
       filteredAttributes: filteredAttrs,
       timestamp: timestamp,
-      value: measurement.value,
+      value: value,
       traceId: traceId,
       spanId: spanId,
     );
--- a/lib/src/metrics/data/exemplar.dart
+++ b/lib/src/metrics/data/exemplar.dart
@@ -43,8 +43,7 @@ class Exemplar {
 
   /// Creates an exemplar from a measurement.
   factory Exemplar.fromMeasurement({
-    required num value,
-    required Attributes measurementAttributes,
+    required Measurement measurement,
     required DateTime timestamp,
     required Attributes aggregationAttributes,
     SpanId? spanId,
@@ -52,7 +51,7 @@ class Exemplar {
   }) {
     // Determine which attributes are filtered out
     final filteredAttrs = _filterAttributes(
-      measurementAttributes,
+      measurement.attributes ?? OTelFactory.otelFactory!.attributes(),
       aggregationAttributes,
     );
 
@@ -60,7 +59,7 @@ class Exemplar {
       attributes: aggregationAttributes,
       filteredAttributes: filteredAttrs,
       timestamp: timestamp,
-      value: value,
+      value: measurement.value,
       traceId: traceId,
       spanId: spanId,
     );
--- a/lib/src/metrics/exemplar_reservoir.dart
+++ b/lib/src/metrics/exemplar_reservoir.dart
@@ -1,0 +1,152 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:math';
+
+import 'package:dartastic_opentelemetry_api/dartastic_opentelemetry_api.dart';
+
+import 'data/exemplar.dart';
+
+/// An ExemplarReservoir receives measurements from instruments and samples them
+/// to provide Exemplars.
+abstract class ExemplarReservoir {
+  /// Offers a measurement to the reservoir.
+  ///
+  /// The reservoir decides whether to retain the measurement as an exemplar.
+  ///
+  /// @param value The value of the measurement.
+  /// @param attributes The attributes associated with the measurement.
+  /// @param context The Context associated with the measurement.
+  /// @param timestamp The time the measurement was recorded.
+  void offerMeasurement(
+      num value, Attributes attributes, Context context, DateTime timestamp);
+
+  /// Returns the currently sampled exemplars and clears the reservoir.
+  ///
+  /// @param pointAttributes The attributes associated with the metric point.
+  /// @return A list of the sampled exemplars.
+  List<Exemplar> collectAndReset(Attributes pointAttributes);
+}
+
+/// A reservoir that retains a fixed number of exemplars.
+///
+/// If more measurements are offered than the fixed size, this implementation
+/// uses a simple random sampling algorithm to decide which exemplars to replace.
+class SimpleFixedSizeExemplarReservoir implements ExemplarReservoir {
+  final int _size;
+  final Random _random = Random();
+  int _measurementsSeen = 0;
+  final List<_MeasurementData?> _storage;
+
+  SimpleFixedSizeExemplarReservoir(this._size)
+      : _storage = List.filled(_size, null);
+
+  @override
+  void offerMeasurement(
+      num value, Attributes attributes, Context context, DateTime timestamp) {
+    int bucket;
+    if (_measurementsSeen < _size) {
+      bucket = _measurementsSeen;
+    } else {
+      bucket = _random.nextInt(_measurementsSeen + 1);
+    }
+
+    if (bucket < _size) {
+      _storage[bucket] = _MeasurementData(
+        value: value,
+        attributes: attributes,
+        context: context,
+        timestamp: timestamp,
+      );
+    }
+    _measurementsSeen++;
+  }
+
+  @override
+  List<Exemplar> collectAndReset(Attributes pointAttributes) {
+    final exemplars = <Exemplar>[];
+    for (var i = 0; i < _storage.length; i++) {
+      final data = _storage[i];
+      if (data != null) {
+        exemplars.add(
+          Exemplar.fromMeasurement(
+            value: data.value,
+            measurementAttributes: data.attributes,
+            timestamp: data.timestamp,
+            aggregationAttributes: pointAttributes,
+            spanId: data.context.spanContext?.spanId,
+            traceId: data.context.spanContext?.traceId,
+          ),
+        );
+        _storage[i] = null;
+      }
+    }
+    _measurementsSeen = 0;
+    return exemplars;
+  }
+}
+
+/// A reservoir that is aligned to histogram buckets.
+///
+/// It keeps the last seen measurement per histogram bucket.
+class AlignedHistogramBucketExemplarReservoir implements ExemplarReservoir {
+  final List<double> _boundaries;
+  final List<_MeasurementData?> _storage;
+
+  AlignedHistogramBucketExemplarReservoir(this._boundaries)
+      : _storage = List.filled(_boundaries.length + 1, null);
+
+  @override
+  void offerMeasurement(
+      num value, Attributes attributes, Context context, DateTime timestamp) {
+    var bucketIndex = _boundaries.length;
+    for (var i = 0; i < _boundaries.length; i++) {
+      if (value <= _boundaries[i]) {
+        bucketIndex = i;
+        break;
+      }
+    }
+    _storage[bucketIndex] = _MeasurementData(
+      value: value,
+      attributes: attributes,
+      context: context,
+      timestamp: timestamp,
+    );
+  }
+
+  @override
+  List<Exemplar> collectAndReset(Attributes pointAttributes) {
+    final exemplars = <Exemplar>[];
+    for (var i = 0; i < _storage.length; i++) {
+      final data = _storage[i];
+      if (data != null) {
+        exemplars.add(
+          Exemplar.fromMeasurement(
+            value: data.value,
+            measurementAttributes: data.attributes,
+            timestamp: data.timestamp,
+            aggregationAttributes: pointAttributes,
+            spanId: data.context.spanContext?.spanId,
+            traceId: data.context.spanContext?.traceId,
+          ),
+        );
+        _storage[i] = null;
+      }
+    }
+    return exemplars;
+  }
+}
+
+class _MeasurementData {
+  final num value;
+  final Attributes attributes;
+  final Context context;
+  final DateTime timestamp;
+
+  _MeasurementData({
+    required this.value,
+    required this.attributes,
+    required this.context,
+    required this.timestamp,
+  });
+}
--- a/lib/src/metrics/exemplar_reservoir.dart
+++ b/lib/src/metrics/exemplar_reservoir.dart
@@ -9,6 +9,10 @@ import 'data/exemplar.dart';
 
 /// An ExemplarReservoir receives measurements from instruments and samples them
 /// to provide Exemplars.
+///
+/// Note for future extension (Issues #151, #153): The reservoir choice should be
+/// user-configurable via a View. Currently, storage constructors hardcode the reservoir
+/// types. Once View parameters are fully wired up, this should be driven by the resolved View.
 abstract class ExemplarReservoir {
   /// Offers a measurement to the reservoir.
   ///
@@ -18,8 +22,10 @@ abstract class ExemplarReservoir {
   /// @param attributes The attributes associated with the measurement.
   /// @param context The Context associated with the measurement.
   /// @param timestamp The time the measurement was recorded.
+  /// @param bucketIndex Optional pre-computed histogram bucket index.
   void offerMeasurement(
-      num value, Attributes attributes, Context context, DateTime timestamp);
+      num value, Attributes attributes, Context context, DateTime timestamp,
+      [int? bucketIndex]);
 
   /// Returns the currently sampled exemplars and clears the reservoir.
   ///
@@ -34,16 +40,18 @@ abstract class ExemplarReservoir {
 /// uses a simple random sampling algorithm to decide which exemplars to replace.
 class SimpleFixedSizeExemplarReservoir implements ExemplarReservoir {
   final int _size;
-  final Random _random = Random();
+  final Random _random;
   int _measurementsSeen = 0;
   final List<_MeasurementData?> _storage;
 
-  SimpleFixedSizeExemplarReservoir(this._size)
-      : _storage = List.filled(_size, null);
+  SimpleFixedSizeExemplarReservoir(this._size, {Random? random})
+      : _random = random ?? Random(),
+        _storage = List.filled(_size, null);
 
   @override
   void offerMeasurement(
-      num value, Attributes attributes, Context context, DateTime timestamp) {
+      num value, Attributes attributes, Context context, DateTime timestamp,
+      [int? bucketIndex]) {
     int bucket;
     if (_measurementsSeen < _size) {
       bucket = _measurementsSeen;
@@ -55,7 +63,8 @@ class SimpleFixedSizeExemplarReservoir implements ExemplarReservoir {
       _storage[bucket] = _MeasurementData(
         value: value,
         attributes: attributes,
-        context: context,
+        traceId: context.spanContext?.traceId,
+        spanId: context.spanContext?.spanId,
         timestamp: timestamp,
       );
     }
@@ -70,12 +79,11 @@ class SimpleFixedSizeExemplarReservoir implements ExemplarReservoir {
       if (data != null) {
         exemplars.add(
           Exemplar.fromMeasurement(
-            value: data.value,
-            measurementAttributes: data.attributes,
+            measurement: OTelAPI.createMeasurement(data.value, data.attributes),
             timestamp: data.timestamp,
             aggregationAttributes: pointAttributes,
-            spanId: data.context.spanContext?.spanId,
-            traceId: data.context.spanContext?.traceId,
+            spanId: data.spanId,
+            traceId: data.traceId,
           ),
         );
         _storage[i] = null;
@@ -88,30 +96,49 @@ class SimpleFixedSizeExemplarReservoir implements ExemplarReservoir {
 
 /// A reservoir that is aligned to histogram buckets.
 ///
-/// It keeps the last seen measurement per histogram bucket.
+/// It MUST store at most one measurement that falls within a histogram bucket,
+/// and SHOULD use a uniformly-weighted sampling algorithm based on the number
+/// of measurements the bucket has seen so far.
 class AlignedHistogramBucketExemplarReservoir implements ExemplarReservoir {
   final List<double> _boundaries;
   final List<_MeasurementData?> _storage;
+  final List<int> _counts;
+  final Random _random;
 
-  AlignedHistogramBucketExemplarReservoir(this._boundaries)
-      : _storage = List.filled(_boundaries.length + 1, null);
+  AlignedHistogramBucketExemplarReservoir(this._boundaries, {Random? random})
+      : _storage = List.filled(_boundaries.length + 1, null),
+        _counts = List.filled(_boundaries.length + 1, 0),
+        _random = random ?? Random();
 
   @override
   void offerMeasurement(
-      num value, Attributes attributes, Context context, DateTime timestamp) {
-    var bucketIndex = _boundaries.length;
-    for (var i = 0; i < _boundaries.length; i++) {
-      if (value <= _boundaries[i]) {
-        bucketIndex = i;
-        break;
+      num value, Attributes attributes, Context context, DateTime timestamp,
+      [int? bucketIndex]) {
+    int index;
+    if (bucketIndex != null) {
+      index = bucketIndex;
+    } else {
+      index = _boundaries.length;
+      for (var i = 0; i < _boundaries.length; i++) {
+        if (value <= _boundaries[i]) {
+          index = i;
+          break;
+        }
       }
     }
-    _storage[bucketIndex] = _MeasurementData(
-      value: value,
-      attributes: attributes,
-      context: context,
-      timestamp: timestamp,
-    );
+
+    final measurementsSeenBucket = _counts[index];
+    if (measurementsSeenBucket == 0 ||
+        _random.nextInt(measurementsSeenBucket + 1) == 0) {
+      _storage[index] = _MeasurementData(
+        value: value,
+        attributes: attributes,
+        traceId: context.spanContext?.traceId,
+        spanId: context.spanContext?.spanId,
+        timestamp: timestamp,
+      );
+    }
+    _counts[index]++;
   }
 
   @override
@@ -122,16 +149,16 @@ class AlignedHistogramBucketExemplarReservoir implements ExemplarReservoir {
       if (data != null) {
         exemplars.add(
           Exemplar.fromMeasurement(
-            value: data.value,
-            measurementAttributes: data.attributes,
+            measurement: OTelAPI.createMeasurement(data.value, data.attributes),
             timestamp: data.timestamp,
             aggregationAttributes: pointAttributes,
-            spanId: data.context.spanContext?.spanId,
-            traceId: data.context.spanContext?.traceId,
+            spanId: data.spanId,
+            traceId: data.traceId,
           ),
         );
         _storage[i] = null;
       }
+      _counts[i] = 0;
     }
     return exemplars;
   }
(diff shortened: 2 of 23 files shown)


lib/src/metrics/data/exemplar.dart
```diff
@@ -43,23 +43,24 @@ class Exemplar {
 
   /// Creates an exemplar from a measurement.
   factory Exemplar.fromMeasurement({
-    required Measurement measurement,
+    required num value,
+    required Attributes measurementAttributes,
     required DateTime timestamp,
```
michaelbushe: The function is "fromMeasurement" but the Measurement parameter is removed?

lib/src/metrics/exemplar_reservoir.dart
```diff
@@ -0,0 +1,152 @@
...
+  final List<_MeasurementData?> _storage;
+
+  SimpleFixedSizeExemplarReservoir(this._size)
+      : _storage = List.filled(_size, null);
+
+  @override
+  void offerMeasurement(
+      num value, Attributes attributes, Context context, DateTime timestamp) {
+    int bucket;
+    if (_measurementsSeen < _size) {
```
michaelbushe, line 56: This is correct, and precisely so.  Nice. I understand that it's the part most implementations get wrong.

`metrics/sdk.md` gives this pseudocode:

```
if num_measurements_seen < num_buckets then
  bucket = num_measurements_seen
else
  bucket = random_integer(0, num_measurements_seen)
end
if bucket < num_buckets then
  reservoir[bucket] = measurement
end
num_measurements_seen += 1
```

Two details you got right that are easy to miss:

1. `random_integer(0, num_measurements_seen)` is **inclusive** on both ends. Dart's `nextInt(n)` returns `[0, n-1]`, so the correct translation is `nextInt(_measurementsSeen + 1)`, which is what you wrote. Using `nextInt(_measurementsSeen)` would bias the reservoir against the most recent measurement forever, and it's the kind of off-by-one that no test catches unless you're specifically testing the distribution.

2. `_measurementsSeen = 0` in `collectAndReset`. The spec requires this: *"Any stateful portion of sampling computation SHOULD be reset every collection cycle."* Without it, after a long-running process has seen a million measurements the replacement probability collapses toward zero, and your exemplars freeze at whatever got sampled in the first minute.

Both right. Good work.

lib/src/metrics/export/metric_config.dart
```diff
@@ -53,6 +54,17 @@ class MetricsConfiguration {
 
     final metricsSdkConfig = MetricsSdkConfig.fromEnvironment();
 
+    ExemplarFilter filter;
+    switch (metricsSdkConfig.exemplarFilter) {
```
michaelbushe, line 58: This is the best decision in the PR..

`MetricsExemplarFilter` (the enum) already existed from #72. It already read `OTEL_METRICS_EXEMPLAR_FILTER`, already had a spec-correct default, and was already threaded down to the exporters. You needed a *behavioural* filter object, which the enum couldn't be.

The tempting move is to add a second config surface: a new env read, a new default, a new public type users have to choose between. Instead you translated the existing enum into the new objects at a single boundary. So there's still exactly one place the env var is read, one default to keep spec-correct, and one thing a user configures.

Had you forked it, we'd now have two sources of truth for one setting, and they'd drift the first time someone changed one and not the other. That class of bug is nearly invisible in review and miserable to diagnose in the field.

**The general principle: when you add an abstraction that overlaps something existing, decide which one is authoritative and make the other a translation of it.** Two independent paths to configure one behaviour is a bug that just hasn't happened yet.

One follow-on now that this switch exists: `MetricsSdkConfig.defaultExemplarFilter` is already `traceBased`, so this switch can never yield a null filter. That makes the nullable field on `MeterProvider` unnecessary. See my comment there.

lib/src/metrics/meter_provider.dart
```diff
@@ -34,6 +35,9 @@ class MeterProvider implements APIMeterProvider {
   /// List of views for configuring metric collection.
   final List<View> _views = [];
 
+  /// The ExemplarFilter used by this provider's meters.
+  ExemplarFilter? exemplarFilter;
```
michaelbushe: This is the one outright spec violation in the PR, and it's a good illustration of how a nullable field quietly encodes a policy decision nobody reviewed.

`metrics/sdk.md`:

> The ExemplarFilter SHOULD be a configuration parameter of a `MeterProvider` for an SDK. **The default value SHOULD be `TraceBased`.**

You got the first half right, it *is* a `MeterProvider` parameter. But the default here is `null`, and both storages guard with `if (exemplarFilter != null && context != null)`. So in practice `null` doesn't mean "unset", it means **AlwaysOff**. The default is off, where the spec says it should be TraceBased.

This isn't theoretical. `MetricsConfiguration.configureMeterProvider` is the only code that ever assigns this field, so any provider it doesn't touch never samples an exemplar. I ran this against your branch:

```dart
final named = OTel.addMeterProvider('named');
print(named.exemplarFilter);            // null
// ... record a counter inside a sampled span ...
// exemplars collected: 0
```

Silently, forever, no warning. Compare the default provider on the same branch, which correctly produced 1.

The fix is one line, make it non-nullable with a real default:

```dart
ExemplarFilter exemplarFilter = const TraceBasedExemplarFilter();
```

Then delete the `!= null` guards in the three storages; they become dead code. `TraceBasedExemplarFilter` already returns `false` when there's no sampled span, so the "no trace context" case is handled by the filter's own logic rather than by the filter's absence.

**The transferable lesson: `null` is not a default, it's the absence of one.** When a spec says a setting has a default, encode that default in the type - a non-nullable field with an initialiser - rather than leaving it nullable and letting each read site decide what null means. Here three separate call sites each independently decided null meant "skip sampling", and that decision was never written down, never reviewed, and contradicts the spec.

Worth knowing for context: this is similar to #270, where named providers export nothing because configuration only ever targets the default provider. Once you've seen the pattern you start noticing it: "who configures this object, and what happens to the ones they miss?" is a good question to ask of any field like this.

Also, the null default breaks a second SHOULD. `metrics/sdk.md` opens the Exemplar section with:

> `Exemplar` sampling SHOULD be turned on by default.

So the null default misses both that and the "default value SHOULD be `TraceBased`" requirement. One nullable field, two SHOULDs, and the observable result is a named provider that silently never samples.

lib/src/metrics/data/exemplar.dart
```diff
@@ -43,23 +43,24 @@ class Exemplar {
 
   /// Creates an exemplar from a measurement.
   factory Exemplar.fromMeasurement({
-    required Measurement measurement,
+    required num value,
```
michaelbushe: The function is "fromMeasurement" but the Measurement parameter is removed?  

After this change, the word `Measurement` appears nowhere in this file except in the factory's own name, and the doc comment directly above still reads *"Creates an exemplar from a measurement."*. So, it should take a Measurement.  Words should flow.

Someone reading `Exemplar.fromMeasurement` reasonably expects to hand it a measurement. Names are a contract with the reader; when the parameters change, the name changes with them. If this shape is what you want, rename it, `Exemplar.sampled(...)` or `Exemplar.fromSample(...)`, and fix the doc comment in the same commit. A stale doc comment is worse than no comment, because a reader trusts it.  

I'm guessing you didn't need to change it, but maybe there's something I'm missing. I'm fairly sure this widened because the reservoir stores `_MeasurementData`, not a `Measurement`, so there was nothing to hand the old factory. But `OTelAPI.createMeasurement<T>(value, attributes)` already exists (`otel_api.dart:770`). The reservoir could build a `Measurement` at collect time and leave this signature untouched.

That would also have avoided the churn in `exemplar_test.dart`, where every call site now repeats:

```dart
value: measurement.value,
measurementAttributes: measurement.attributes ?? OTelFactory.otelFactory!.attributes(),
```

That null-fallback used to live *once*, inside this factory. Now it's duplicated at every call site, and each copy is an opportunity to get it wrong.  Always be [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). 

Always thing of the user, the method caller, and serve them (and never yourself).  **When a change makes callers repeat logic the callee used to own, that's usually a sign the change is pointed the wrong way.** It's a useful smell to watch for: if your diff adds the same two lines to six call sites, ask whether the thing you changed should have absorbed them instead.

**This is a breaking change to exported public API.** `exemplar.dart` is exported from `dartastic_opentelemetry.dart`, so any user calling `Exemplar.fromMeasurement(measurement: m, ...)` fails to compile after this lands. The CHANGELOG entry only mentions that exemplars were implemented. Breaking changes need their own line. That entry is the only thing standing between a user and an unexplained build failure, and it's the one part of a CHANGELOG that isn't optional.

Even though we are still in beta, we get tens of thousands downloads a week and need to prove to user, all of whom are developing new trust in us, that we won't break them, we respect their time and respect their systems.

lib/src/metrics/exemplar_reservoir.dart
```diff
@@ -0,0 +1,152 @@
...
+  List<Exemplar> collectAndReset(Attributes pointAttributes);
+}
+
+/// A reservoir that retains a fixed number of exemplars.
+///
+/// If more measurements are offered than the fixed size, this implementation
+/// uses a simple random sampling algorithm to decide which exemplars to replace.
+class SimpleFixedSizeExemplarReservoir implements ExemplarReservoir {
+  final int _size;
+  final Random _random = Random();
```
michaelbushe: `Random()` here is unseeded and can't be replaced from outside, which makes this class's central behaviour - the sampling distribution - untestable.

I think this is the *root cause* of the missing tests rather than a separate issue, and that's why I'm flagging it as the most important thing in this review after the null default. You can't write "offer 1000 measurements to a size-1 reservoir and assert every one had a fair chance" against an unseeded generator. So the test doesn't get written. So the algorithm ships unverified. It happens to be correct here. I checked it line by line against the spec. Next time it might not be, and nothing would tell you.

The fix is small and costs nothing in production:

```dart
final Random _random;

SimpleFixedSizeExemplarReservoir(this._size, {Random? random})
    : _storage = List.filled(_size, null),
      _random = random ?? Random();
```

Now a test passes `Random(42)`, gets a deterministic sequence, and can assert exactly which slots were filled after a known series of offers. Production behaviour is identical.

**The general principle, and it's the most useful thing I can give you here: when code is hard to test, that's usually the design telling you something, not a reason to skip the test.**

The three most common versions of this, all with the same fix:

- unseeded randomness (this)
- `DateTime.now()` read from deep inside a class (the storages do this too, passing the timestamp into `record()` would let a test assert exemplar timestamps)
- constructing collaborators directly instead of accepting them

In each case: let the caller inject it, default to the real thing. You keep the convenient API and gain a testable one.

A good habit to build is to try writing the test first for anything with probabilistic or time-dependent behavior. Not out of TDD dogma, but because the moment you sit down to write that test you'll discover this problem in thirty seconds, while it's still one line to fix.

lib/src/metrics/exemplar_reservoir.dart
```diff
@@ -0,0 +1,152 @@
...
+      }
+    }
+    return exemplars;
+  }
+}
+
+class _MeasurementData {
+  final num value;
+  final Attributes attributes;
+  final Context context;
```
michaelbushe: Storing the whole `Context` here misses two SHOULDs in `metrics/sdk.md`, and since we implement every SHOULD rather than only the MUSTs, both need addressing.

The first is about this exact decision:

> The "offer" method SHOULD have the ability to pull associated trace and span
> information without needing to record full context. In other words, current
> span context and baggage can be inspected at this point.

That is precisely the shape being recommended: inspect the context at offer time, extract what you need, and do not retain the context itself. This implementation does the opposite, storing the full `Context` in `_MeasurementData` and deferring extraction to `collectAndReset`.

The second follows from it:

> The `ExemplarReservoir` SHOULD avoid allocations when sampling exemplars.

Every accepted offer allocates a `_MeasurementData`. Once you are storing plain values rather than a live context, the slots become reusable, and the allocation per sampled measurement can go away entirely.

There is a concrete cost too, not just a spec citation. A `Context` transitively reaches the active `Span`, which holds its attributes, events, and links. Every sampled measurement therefore pins that whole object graph alive until the next collection cycle, up to 60 seconds by default. For a histogram with 20 buckets across a few hundred attribute sets, that is a lot of spans kept reachable long
after they ended and were exported. You only ever use two fields from it, once:

```dart
spanId: data.context.spanContext?.spanId,
traceId: data.context.spanContext?.traceId,
```

So extract them at offer time and store only what you need:

```dart
class _MeasurementData {
  final num value;
  final Attributes attributes;
  final TraceId? traceId;
  final SpanId? spanId;
  final DateTime timestamp;
}
```

The general principle, which is worth more than this instance: retain the narrowest thing that satisfies the need, and narrow it at the moment of capture rather than the moment of use. Holding a broad handle "in case we need more later" is how caches quietly become leaks. Nothing looks wrong at the call site, and the cost surfaces only as steadily climbing memory in a long running process,
which is one of the harder things to trace back to its cause.

A secondary benefit: after this change `_MeasurementData` is plainly immutable data with no reference to anything live, which makes it far easier to construct in a test.

lib/src/metrics/exemplar_reservoir.dart
```diff
@@ -0,0 +1,152 @@
...
+    }
+    _measurementsSeen = 0;
+    return exemplars;
+  }
+}
+
+/// A reservoir that is aligned to histogram buckets.
+///
+/// It keeps the last seen measurement per histogram bucket.
+class AlignedHistogramBucketExemplarReservoir implements ExemplarReservoir {
```
michaelbushe, line 102: Keeping the last measurement per bucket is permitted by the spec, but it is not what we ship. `metrics/sdk.md` says this reservoir "MUST store at most one measurement that falls within a histogram bucket, and SHOULD use a uniformly-weighted sampling algorithm based on the number of measurements the bucket has seen so far," and only then offers the alternative: "the implementation MAY instead keep the last seen measurement."

This project implements every SHOULD in the specification as a matter of policy, not just the MUSTs. That rule exists because it's a new implementation - we don't have a legacy to honor so it's easier, and we are asking people to trust this SDK as a faithful implementation, and "technically permitted" is a much weaker promise than "does what the spec recommends." So when the spec offers a SHOULD and a MAY, we take the SHOULD unless there is a documented reason we cannot, and the reason has to be better than "the other one was simpler."  We haven't had a reason not to implement a SHOULD yet.

The spec even gives you the algorithm:

```
bucket = find_histogram_bucket(measurement)
num_measurements_seen_bucket = num_measurements_seen[bucket]
if random_integer(0, num_measurements_seen_bucket) == 0 then
  reservoir[bucket] = measurement
end
num_measurements_seen[bucket] += 1
```

So `AlignedHistogramBucketExemplarReservoir` needs a per-bucket counter alongside `_storage`, and the same reset discipline you already applied correctly in `SimpleFixedSizeExemplarReservoir`: `metrics/sdk.md` requires that "any stateful portion of sampling computation SHOULD be reset every collection cycle," which now means clearing the per-bucket counts in `collectAndReset`, not just the
slots. You have already shown you understand that requirement, this is the same idea applied per bucket.

Two smaller things in the same class.

The reservoir is always chosen for histograms regardless of bucket count, but the spec scopes it: "Explicit bucket histogram aggregation with more than 1 bucket SHOULD use `AlignedHistogramBucketExemplarReservoir`," and "all other aggregations SHOULD use `SimpleFixedSizeExemplarReservoir`." A histogram configured with a single boundary should fall to the simple reservoir.

And `offerMeasurement` re-derives the bucket index with a linear scan, when `HistogramStorage.record` has already computed the same bucket for the same value moments earlier. Pass it through rather than computing it twice on every recorded measurement. On a hot histogram that is real cost for no benefit.

lib/src/metrics/storage/sum_storage.dart
```diff
@@ -56,16 +62,33 @@ class SumStorage<T extends num> extends NumericStorage<T> {
...
+      pointData = _SumPointData<T>(
         value: value,
         lastUpdateTime: DateTime.now(),
+        reservoir:
+            SimpleFixedSizeExemplarReservoir(1), // Simple fixed size of 1
       );
+      _points[attributes] = pointData;
+    }
+
+    // Process exemplars if a filter is provided
```
michaelbushe: This block is copied verbatim into three files: here, `gauge_storage.dart:37`, and `histogram_storage.dart:58`. All three null-check the filter, null-check the context, call `shouldSample`, then `offerMeasurement` with the same four arguments and the same `attributes ?? OTelFactory.otelFactory!.attributes()` fallback.

There are already shared base types for exactly this in `metric_storage.dart` - `MetricStorage`, `NumericStorage<T>`, `HistogramStorageBase<T>`. A mixin fits well here, since the three storages have different point types but identical sampling policy:

```dart
mixin ExemplarSampling<T extends num> {
  ExemplarFilter get exemplarFilter;

  void maybeOffer(ExemplarReservoir reservoir, T value,
      Attributes attributes, Context context, DateTime timestamp) {
    if (exemplarFilter.shouldSample(value, attributes, context)) {
      reservoir.offerMeasurement(value, attributes, context, timestamp);
    }
  }
}
```

Each storage then calls `maybeOffer(...)` and the policy lives in exactly one place.

**Why this matters more than eight duplicated lines might suggest:** every other fix in this review has to be applied three times. Making the filter non-nullable removes a guard here, three times. Injecting the timestamp for testability changes this call three times. Narrowing what gets stored changes the arguments three times.

Three copies means three chances to fix two and forget the third. And nothing will catch it, because there are no tests. That's how you end up with exemplars that work for counters and silently don't for gauges, and nobody notices for two releases.

The rule I'd offer: **duplication is cheap to write and expensive to change, so the useful question isn't "is this duplicated?" but "when this changes, how many places have to change together, and will anything tell me if I miss one?"** Three call sites that must always agree, with no test holding them together, is a clear trigger to extract , even when each copy looks small enough to ignore.

lib/src/metrics/storage/gauge_storage.dart
```diff
@@ -3,26 +3,44 @@
...
   @override
-  void record(T value, [Attributes? attributes]) {
+  void record(T value, [Attributes? attributes, Context? context]) {
     // Create a normalized key for lookup
     final key = attributes ?? _emptyAttributes();
 
+    final pointData = _GaugePointData<T>(
+      value: value,
+      updateTime: DateTime.now(),
+      reservoir: SimpleFixedSizeExemplarReservoir(1), // Simple fixed size of 1
```
michaelbushe: This is a real bug, and it only affects gauges. A new reservoir is constructed on every single `record()` call, then `_points[key] = pointData` replaces the previous point wholesale. So any exemplar sampled earlier in the collection interval is thrown away by the next measurement.

Compare `SumStorage`, which gets this right:

```dart
if (_points.containsKey(attributes)) {
  pointData = _points[attributes]!;      // reuses the existing reservoir
  pointData.add(value);
} else {
  pointData = _SumPointData<T>(reservoir: SimpleFixedSizeExemplarReservoir(1), ...);
  _points[attributes] = pointData;
}
```

`HistogramStorage` reuses the point's reservoir too. Gauge is the one that diverged.

I ran the identical sequence against both, recording once inside a sampled span and then once outside any span:

```
GAUGE metric=g value=7.0 exemplars=0     the span's exemplar was destroyed
SUM   metric=c value=12  exemplars=1     correct
```

The first measurement is accepted by `TraceBasedExemplarFilter` and lands in the reservoir. The second arrives with no active span, so the filter rejects it, but by then the fresh `_GaugePointData` has already replaced the old one and taken its reservoir with it. Collection returns nothing, when it should return the exemplar from the sampled span.

For last value semantics, replacing the *value* on every record is correct. Replacing the *reservoir* is not. The reservoir is supposed to accumulate across the whole collection interval, and `collectAndReset` is the thing that clears it. The spec is explicit at `metrics/sdk.md`: "A new `ExemplarReservoir` MUST be created for every known timeseries data point." Per data point, not per measurement.

The fix mirrors `SumStorage`: look up the existing point, and if it exists replace only `value` and `updateTime` while keeping its reservoir.

Two things worth taking from this beyond the fix itself.

First, this is the concrete version of the duplication point I made on `sum_storage.dart`. Three hand maintained copies of the same logic, and one of them quietly diverged. It is not that somebody forgot to apply a fix, it is that the three were never the same to begin with, and nothing in the codebase was capable of noticing.

Second, and more important: this shipped with green CI. Every check passed. That is worth sitting with, because it is the clearest possible demonstration that green CI means "nothing that was already tested broke," and nothing more. A single test asserting that a gauge recorded inside a sampled span still reports an exemplar after a later unsampled record would have caught this in seconds. That test does not exist, so the bug was invisible.

lib/src/metrics/exemplar_reservoir.dart
```diff
@@ -0,0 +1,152 @@
...
+
+import 'dart:math';
+
+import 'package:dartastic_opentelemetry_api/dartastic_opentelemetry_api.dart';
+
+import 'data/exemplar.dart';
+
+/// An ExemplarReservoir receives measurements from instruments and samples them
+/// to provide Exemplars.
+abstract class ExemplarReservoir {
```
michaelbushe, line 16: One requirement this interface does not yet satisfy, noted here because it shapes the design rather than any single line.

`metrics/sdk.md`, Custom ExemplarReservoir:

> The SDK MUST provide a mechanism for SDK users to provide their own ExemplarReservoir implementation. This extension MUST be configurable on a metric View, although individual reservoirs MUST still be instantiated per metric-timeseries.

Right now the reservoir type is hardcoded inside each storage constructor, so there is no seam for a user to supply their own. That is two MUSTs unmet: the mechanism itself, and its configurability on a View.

I want to be fair about this, because it is not really fixable inside this PR. `View` on main is currently vestigial: `MeterProvider` collects views through `addView()` and nothing ever reads them to configure aggregation. That is issue #151, already open, with #153 covering the missing stream parameters. So you are not breaking a working extension point, you are building where one will have to exist.

What that means practically: when #151 and #153 land, the reservoir choice has to move from "hardcoded in the storage constructor" to "supplied by the resolved view, defaulting to the built in choice for the aggregation." Worth a comment in the code now recording that, so whoever picks up #153 finds the connection instead of rediscovering it.

The general habit here is worth naming. When you implement a feature that the spec says must be user extensible, and the extension point is blocked on other work, say so in the code and link the issue. Silence reads as "nobody thought about it," which costs the next person the whole investigation you already did.

michaelbushe (commented):
Thanks for this, this is huge and not easy. What you got right, because it's the majority of the PR and some of it is the hard part:

**The reservoir algorithm is correct.** `SimpleFixedSizeExemplarReservoir` is a faithful implementation of the pseudocode in `metrics/sdk.md` (Reservoir sampling), including the two details people usually miss: the random bound is *inclusive* of the current count, and `_measurementsSeen` resets on collection. I verified both against the spec text.

**You put the reservoir in the right layer.** Per-point reservoir state belongs next to the per-point aggregation state it parallels, which is exactly where you put it. Threading `Context` through `record()` is the minimal correct change, and capturing `Context.current` in the *instrument* at record time (rather than at collection time) is right and a good insight. Capturing later would attach whichever span happened to be active during export, which would be meaningless.

**You reconciled with the existing config surface** instead of forking a second one. See my note on `metric_config.dart`. That was the most consequential design decision in this PR and you made the right call.

I also confirmed it works end to end. I ran three probes on your branch:

| probe | filter resolved | exemplars produced |
|---|---|---|
| `OTel.initialize()`, counter inside a sampled span | `TraceBasedExemplarFilter` | 1 |
| `OTEL_METRICS_EXEMPLAR_FILTER=always_on`, no span | `AlwaysOnExemplarFilter` | 1 |
| `OTel.addMeterProvider('named')` | **`null`** | **0** |

That third row is the one real spec deviation, and it's detailed inline.

---

**The thing to take from this review, above any individual fix: this PR adds 203 lines of sampling logic and zero tests for it.**

There's a puzzle that's often told - You are a CEO of a half-a-billion dollar company.  You have a great engineering team and your code is fully tested. A hacker calls you and says, "I have access to all your code. If you don't give a billion dollars, I'm going to delete either your tests or your code."

You say, "I don't have a billion dollars."

He asks, "Then which is it, your tests, or your code?"

Even before AI, the right answer is "the code". Because if you have to write the code again, at least you know what the target it.  If you delete the tests, you'll never know if the next change is correct.  (With AI, the answer is more obvious because the tests are the spec and AI can run them until they all pass.  Same for humans but faster.)

Make sure you run ./tool/coverage.sh before you submit. If the coverage goes down, then .  I'm trying to make things fail if this ever happens but apparently, that's not working yet.  It should fail on push if you set up the git hooks in the repo, see CONTRIBUTING.md.  We have 100% (or 99.9?) coverage on the API, 93% on the SDK and I want it to go UP and never down.  Shoot for that. 

The only test change is mechanical: call sites unpacking `measurement.value` and `measurement.attributes` because a signature changed. The 97 lines you deleted were legitimately dead (`addExemplar` is genuinely gone from the interface), but nothing replaced them. Net test delta is -79 lines.

Right now, without the working coverage check, green CI on a PR like this means "nothing that was already tested broke." It does not mean the new code works, and it means less if you delete tests. 

Three separate defects flagged below would have been caught by tests you'd naturally write while building this:

- A test asserting a named `MeterProvider` produces exemplars → catches the null-default bug.
- A test offering 1000 measurements to a size-1 reservoir and asserting the distribution → impossible today, because `Random` isn't injectable. That's a design problem the test would have surfaced *before* you shipped it.
- A test asserting a sampled exemplar carries the right `traceId`/`spanId` → would have made you look at what `_MeasurementData` retains.

That's the general lesson: **when code is hard to test, that's usually the code telling you something about its design, not a reason to skip the test.** The unseeded `Random` is the clearest example here.

For calibration, our internal reference implementation of this same issue carries ~350 lines of tests against ~250 lines of implementation. That ratio isn't a rule, but for sampling logic, where the whole point is probabilistic behavior you can't eyeball, it's the right order of magnitude.

michaelbushe:
This is awesome @harshitt13 !  Higher coverage than previously. I think that record() gained an optional DateTime? timestamp, the context ?? Context.current, and the bucketIndex pass-through so the histogram bucket isn't computed twice.  Exemplars are here!

Issue #281, opened 2026-08-27
[BUG] OTLP gRPC: a URL-form endpoint without a port never gets the default 4317 appended


michaelbushe:
## Summary

`OtlpGrpcExporterConfig._validateEndpoint` intends to append the default gRPC port to a URL-form endpoint that has none, but the branch is unreachable, so it never happens.

## Offending code

`lib/src/trace/export/otlp/otlp_grpc_span_exporter_config.dart:126-129`

```dart
if (uri.port == 0 && !endpoint.contains(':')) {
  // No port specified in URL format, add default
  return '${uri.scheme}://${uri.host}:4317${uri.path}';
}
```

This branch only runs for an endpoint containing no colon at all. Any URL-form endpoint contains `://`, so `!endpoint.contains(':')` is always false here and the two lines are dead code.

## Observed behaviour

```
http://collector.example.com   ->  http://collector.example.com      (no port added)
https://c.example.com/ingest   ->  https://c.example.com/ingest      (no port added)
collector.example.com          ->  collector.example.com:4317        (correct, different path)
http://c.example.com:4317      ->  http://c.example.com:4317         (correct)
```

The bare-host path works. Only the URL-form path is affected.

## Why it matters

`protocol/exporter.md` gives the OTLP/gRPC default endpoint as `http://localhost:4317`, and the endpoint option "MUST accept a URL with a scheme of either `http` or `https`". A user who sets a scheme'd endpoint without a port, which is a natural thing to write, gets no port at all, and the gRPC client then falls back to its own default rather than 4317.

## Suggested fix

Drop the `!endpoint.contains(':')` guard. `uri.port == 0` already means "no port was specified", since `Uri.port` returns 0 when no port is present and no scheme default applies:

```dart
if (uri.port == 0) {
  return '${uri.scheme}://${uri.host}:4317${uri.path}';
}
```

The following block, which validates a malformed port, is guarded on `uri.port == 0 && endpoint.contains(':')` and would need reordering or folding into the same branch.

## Coverage note

These lines show as uncovered in `coverage/lcov.info`, which is how this was found. A test asserting `http://host` becomes `http://host:4317` will stay red until this is fixed.

PR #282, opened 2026-08-27
test(metrics): named-provider exemplar regression, reader failure paths, endpoint validation

michaelbushe:
Follow-up to #277. Adds the regression test that PR was missing, and picks off three uncovered areas found in `coverage/lcov.info`.

Tests only, plus one export fix described below. 15 tests added, 14 green and 1 skipped against a filed issue.

## Named-provider exemplars (4 tests)

#277 fixed the null exemplar filter default, but only the gauge half of that work got a regression test. The null default was found by probing rather than by the suite, so nothing stops it coming back. These pin that a provider from `addMeterProvider` defaults to `TraceBased`, samples inside a sampled span, and still honours an explicitly assigned `AlwaysOff`.

## Export fix

Writing that test surfaced a real gap. `ExemplarFilter` and `ExemplarReservoir` were never exported:

```dart
OTel.initialize(exemplarFilter: ...)      // takes one
MeterProvider.exemplarFilter              // public field
export 'src/metrics/exemplar_filter.dart' // missing
```

So an application could not construct a filter to pass, or even name the type of `meterProvider.exemplarFilter`. The programmatic configuration path the spec requires was unreachable. Both files are now exported.

This also removed four `unnecessary_import` warnings, two of them in tests that had been importing `package:dartastic_opentelemetry/src/metrics/...` directly to work around the omission. Reaching into `src/` from a test is a reliable signal that something should have been exported.

## PeriodicExportingMetricReader (5 tests)

Covers the `interval` and `timeout` accessors and the `forceFlush` and `shutdown` failure paths, pinning that an exporter which throws is reported as `false` rather than propagating to the caller.

## OtlpGrpcExporterConfig endpoint validation (6 tests + 1 skipped)

Covers default-port appending for a bare host, explicit ports in both forms, a malformed port, and an empty host.

The skipped case documents **#281**. The branch meant to append `:4317` to a URL-form endpoint is guarded on `!endpoint.contains(':')`, which no URL-form endpoint can satisfy because of `://`. It is dead code, and `http://collector.example.com` keeps no port at all:

```
http://collector.example.com  ->  http://collector.example.com   (should be :4317)
collector.example.com         ->  collector.example.com:4317     (correct, other path)
```

The test is written to pass once #281 is fixed, so un-skipping it is the acceptance criterion.

## Verification

- `dart analyze`: no issues
- `dart test test/unit/`: 2183 pass, 4 skipped, 0 failures
- Full `tool/coverage.sh` baseline before this PR was 93.5%, 5249 of 5614 lines

Relates to #277, #281.

### DIFF
--- a/test/unit/metrics/named_provider_exemplar_test.dart
+++ b/test/unit/metrics/named_provider_exemplar_test.dart
@@ -1,0 +1,82 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Regression test for the exemplar filter default on providers that
+// MetricsConfiguration never configures.
+//
+// While MeterProvider.exemplarFilter was nullable and both storages guarded
+// on `exemplarFilter != null`, null meant "sample nothing". Only the default
+// provider is assigned a filter, by MetricsConfiguration.configureMeterProvider,
+// so any provider from addMeterProvider silently produced no exemplars at all.
+//
+// metrics/sdk.md requires the opposite: "Exemplar sampling SHOULD be turned on
+// by default" and "The ExemplarFilter ... default value SHOULD be TraceBased".
+
+import 'package:dartastic_opentelemetry/dartastic_opentelemetry.dart';
+import 'package:test/test.dart';
+
+void main() {
+  setUp(() async {
+    await OTel.reset();
+    EnvironmentService.testOverrides = {'OTEL_TRACES_EXPORTER': 'none'};
+    await OTel.initialize(
+      serviceName: 'named-provider-exemplar-test',
+      detectPlatformResources: false,
+      enableLogs: false,
+    );
+  });
+
+  tearDown(() async {
+    try {
+      await OTel.shutdown();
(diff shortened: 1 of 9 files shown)


test/unit/metrics/named_provider_exemplar_test.dart
```diff
@@ -0,0 +1,82 @@
...
+
+  tearDown(() async {
+    try {
+      await OTel.shutdown();
+    } catch (_) {}
+    await OTel.reset();
+    EnvironmentService.testOverrides = null;
+  });
+
+  /// Records [value] on a counter from [provider] inside a sampled span and
```
harshitt13, line 37: The dartdoc says `Records [value] on a counter...`, but the helper function doesn't actually take a `value` parameter (it hardcodes `counter.add(7)`). The `[value]` syntax will result in a broken doc reference.

harshitt13 (approved):
LGTM! Just noticed one tiny thing

michaelbushe:

PR #287, opened 2026-08-27
fix(tool): correct the script paths and make coverage_unit_only real

michaelbushe:
Three problems in the `scripts:` block, each of which meant a command did not do what its name said.

## `coverage` was broken

```yaml
coverage: ./tools/coverage.sh    # the directory is tool/
```

The command failed outright in both packages.

## `coverage_unit_only` ran nothing

```yaml
coverage_unit_only: dart test --tags=unit --coverage=coverage
```

Nothing is tagged. The API has no `@Tags` anywhere and no `dart_test.yaml`; the SDK has exactly one tagged file out of 157. Running it in the API:

```
No tests ran.
No tests match the requested tag selectors:
  include: "unit"
```

It exited successfully having tested nothing, which is worse than failing.

The unit/integration split in both packages is by **directory**, not tags: `test/unit/` holds 93 files in the API and 157 in the SDK. So `coverage.sh` now takes `--unit` and restricts the run to `test/unit`. Both packages accept the same flag, so all three commands mean the same thing in each.

## `test` had drifted

`dart test` in the SDK, a script in the API. Both now use `./tool/test.sh`.

## Also

The API's `coverage.sh` gains the `darwin*` guard around `open` that the SDK already had, with the same comment. Unguarded it fails on Linux, and the SDK runs `coverage.sh` in CI under `set -e`, so an unguarded `open` there would fail the build.

The SDK's `coverage.sh` keeps `--log`, `--concurrency` and `--fail-fast`; the API's had no argument parsing at all, so it gains a parser with just `--unit`. That flag is a no-op in the API today, since `test/unit` is the only directory, but the interface matches.

## Verification

`bash -n` clean on both scripts. Flag parsing checked: `--unit` selects `./test/unit`, no flag selects `./test`, and an unknown flag still errors with usage.


michaelbushe:

PR #298, opened 2026-09-04
fix(sampling): correct TraceIdRatioBased sampler description

### LINKED ISSUE #127: [BUG] Trace SDK: TraceIdRatioBased sampler description has the wrong form
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Trace SDK, TraceIdRatioBased
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/trace/sdk/#traceidratiobased
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/sdk.md#traceidratiobased

> * Description MUST return a string of the form `"TraceIdRatioBased{RATIO}"` with `RATIO` replaced with the Sampler instance's trace sampling ratio represented as a decimal number.

## The problem

The sampler returns `TraceIdRatioSampler{...}`. The spec requires `TraceIdRatioBased{RATIO}`.

The word `Based` is missing. Any tool that reads the sampler description gets a name that no other OpenTelemetry SDK produces.

## Offending code

`dartastic_opentelemetry/lib/src/trace/sampling/trace_id_ratio_sampler.dart:18`

```dart
@override
String get description => 'TraceIdRatioSampler{$ratio}';
```

## How to correct it

Return the required form. Write the ratio as a decimal number:

```dart
String get description => 'TraceIdRatioBased{${ratio.toStringAsFixed(6)}}';
```

## Notes

- Found by: Fable 5 (both runs), Grok 4.5, GPT-5.6 Sol
- Source finding files: `fable-ultra/trace-sdk-08-traceidratio-description.md`, `fable-medium/trace-sdk-traceidratio-description.md`, `grok/trace/012-trace-id-ratio-description.md`, `codex/trace-ratio-sampler-description.md`
- Parent issue: #95

LaPyaeOo:
## Summary

* Correct `TraceIdRatioSampler` description to use the `TraceIdRatioBased{RATIO}` format.
* Format the ratio with 6 decimal places.
* Update the corresponding test.

## Verification

* `dart format` clean
* `dart analyze` clean
* `dart test test/unit/trace/sampling/trace_id_ratio_sampler_test.dart` passed

Fixes #127

### DIFF
--- a/lib/src/trace/sampling/trace_id_ratio_sampler.dart
+++ b/lib/src/trace/sampling/trace_id_ratio_sampler.dart
@@ -15,7 +15,7 @@ class TraceIdRatioSampler implements Sampler {
   final double ratio;
 
   @override
-  String get description => 'TraceIdRatioSampler{$ratio}';
+  String get description => 'TraceIdRatioBased{${ratio.toStringAsFixed(6)}}';
 
   /// Creates a TraceIdRatioSampler with the given ratio.
   /// [ratio] must be in the range [0.0, 1.0].
--- a/test/unit/trace/sampling/trace_id_ratio_sampler_test.dart
+++ b/test/unit/trace/sampling/trace_id_ratio_sampler_test.dart
@@ -25,7 +25,7 @@ void main() {
 
     test('description returns expected value', () {
       final sampler = TraceIdRatioSampler(0.5);
-      expect(sampler.description, equals('TraceIdRatioSampler{0.5}'));
+      expect(sampler.description, equals('TraceIdRatioBased{0.500000}'));
     });
 
     test('sampler with ratio 0.0 never samples', () {


lib/src/trace/sampling/trace_id_ratio_sampler.dart
```diff
@@ -15,7 +15,7 @@ class TraceIdRatioSampler implements Sampler {
   final double ratio;
 
   @override
-  String get description => 'TraceIdRatioSampler{$ratio}';
+  String get description => 'TraceIdRatioBased{${ratio.toStringAsFixed(6)}}';
```
robert-northmind: The name correction looks right, but `toStringAsFixed(6)` loses precision. E.g. `0.1234561` and `0.1234562` both display as `0.123456`, and `0.0000001` looks like zero.

I see this follows the suggestion in #127, but the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/sdk.md#traceidratiobased) recommends enough precision to distinguish different ratios. 
I think we can keep Dart’s existing formatting and just correct the name:

```dart
String get description => 'TraceIdRatioBased{$ratio}';
```

I ran some local tests and reproduced the precision loss and passed with this getter, alongside the existing sampler tests. 
Should we add regression tests for these ratios too maybe?

LaPyaeOo: I have already fixed not to lose precision for distinct or very small ratios. Either in the test file.

robert-northmind (approved):
Nice. Looks good 👍

PR #303, opened 2026-09-10
The OTLP log exporters send event_name

michaelbushe:
#### Description

`LogRecord.eventName` was never copied onto the wire. Every record emitted with `emit(eventName: ...)` reached the collector as a plain log record with no name, so anything filtering on `event_name` saw nothing. Attributes, body, scope and resource were all intact; only that field was missing.

The transformer could not have set it: the bundled protobuf definitions were generated from opentelemetry-proto **v1.1.0**, which predates `LogRecord.event_name` (field 12, added in v1.4.0). `tool/setup_protos.sh` now pins **v1.11.0** and the generated files are regenerated. That also brings `EntityRef` on `Resource` and the string-table fields on `AnyValue` and `KeyValue`; no generated type was removed and no other exporter code changes. `_transformLogRecord` copies `eventName` when present and leaves the field unset otherwise.

`protos/` (where the generator clones upstream) is now gitignored so a regeneration does not offer the clone for commit.

#### Link to tracking issue
None filed; found from a wire capture.

#### Testing

`log_record_transformer_test.dart`:
- the existing single-record test now asserts `eventName` on the transformed record (it already passed `eventName: 'test.event'` in and never checked it);
- a new test round-trips through `writeToBuffer()` / `fromBuffer()` and asserts the name survives the wire encoding, and that it is the `event_name` field rather than an `event.name` attribute;
- a new test asserts a record with no event name leaves the field unset.

The first two fail before the change (the generated class had no `eventName` at all). Also added `otlp_http_log_record_exporter_config_test.dart` for the endpoint normaliser's two refusal paths, which had no coverage. Full suite and coverage pass locally via the pre-push hook (93.65%).

#### CHANGELOG
- [x] `CHANGELOG.md` updated under 1.1.0-beta.16-wip (Fixed). No PR link in the entry yet; add it if you want the number there.

#### Documentation
None needed; the API already documents `eventName` on `LogRecord` and `emit`.

#### Spec Compliance
Logs Data Model, field `EventName`: https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-eventname ("Type: string ... Event name identifies the class or type of the Event"). OTLP carries it as `LogRecord.event_name` since opentelemetry-proto v1.4.0. The OTLP exporter MUST transmit the log record's fields; it was silently dropping this one.

#### Authorship
- [ ] I, a human, authored this pull request and stand behind these changes.
- Assisted-by: Claude Fable 5.1

https://claude.ai/code/session_01V7j6yPYdv4SXz2Ti5x8hLS

### DIFF
--- a/tool/setup_protos.sh
+++ b/tool/setup_protos.sh
@@ -5,7 +5,7 @@ set -e
 PROTO_DIR="protos"
 OUTPUT_DIR="lib/proto"
 TEMP_DIR=".proto_gen_temp"
-OPENTELEMETRY_PROTO_VERSION="v1.1.0"  # Update this to the version you want to use
+OPENTELEMETRY_PROTO_VERSION="v1.11.0"  # Update this to the version you want to use
 
 # Create directories if they don't exist
 mkdir -p "$PROTO_DIR/opentelemetry-proto"
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@
 /.run/
 /dartastic_opentelemetry.iml
 /.claude/settings.local.json
+
+# Cloned by tool/setup_protos.sh; generated output lives in lib/proto
+protos/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- **The OTLP log exporters now send `event_name`.** `LogRecord.eventName` was never copied onto the wire, so every
+  event emitted with `emit(eventName: ...)` arrived at the collector as a plain log record with no name, and anything
+  filtering on `event_name` saw nothing. The bundled protobuf definitions were generated from opentelemetry-proto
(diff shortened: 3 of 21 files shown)


tool/setup_protos.sh
```diff
@@ -5,7 +5,7 @@ set -e
 PROTO_DIR="protos"
 OUTPUT_DIR="lib/proto"
 TEMP_DIR=".proto_gen_temp"
-OPENTELEMETRY_PROTO_VERSION="v1.1.0"  # Update this to the version you want to use
+OPENTELEMETRY_PROTO_VERSION="v1.11.0"  # Update this to the version you want to use
```
robert-northmind, line 8: Should we also update the existing proto checkout when this version changes? If protos/opentelemetry-proto/.git is already there from v1.1.0, the clone block is skipped and we regenerate the old schema. That removes LogRecord.eventName, so the new transformer no longer compiles.

We could fetch/check out the requested tag before regenerating, and verify the upgrade with a v1.1.0 cache already present.

robert-northmind (approved):
Looks good. just left one small comment

PR #305, opened 2026-09-12
fix: OTLP HTTP retryable status codes

### LINKED ISSUE #224: [BUG] OTLP: HTTP exporters do not retry 502 and 504
## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — OTLP Exporter, Retry and Transient errors
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/protocol/exporter/#retry
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/protocol/exporter.md#retry

> Transient errors MUST be handled with a retry strategy. This retry strategy MUST implement an exponential back-off with jitter to avoid overwhelming the destination until the network is restored or the destination has recovered.

> For OTLP/HTTP, transient errors are defined by: 1. A set of retryable HTTP status codes received from the server.

The OTLP protocol specification lists that set as 429, 502, 503, and 504.

## The problem

All three OTLP/HTTP exporters treat two status codes as retryable.

A 502 and a 504 both sit in the retryable set. This code fails permanently on either one. Telemetry behind a proxy or a load balancer that fails for a moment is dropped instead of retried.

A second weakness sits in the same path. The code decides retryability by matching a substring of an exception message, not by reading the status code.

## Offending code

`dartastic_opentelemetry/lib/src/trace/export/otlp/http/otlp_http_span_exporter.dart:25-28`

```dart
static const _retryableStatusCodes = [
  429, // Too Many Requests
  503, // Service Unavailable
];
```

The same list is at `metrics/export/otlp/http/otlp_http_metric_exporter.dart:19-21` and `logs/export/otlp/http/otlp_http_log_record_exporter.dart:24-26`. The substring match is at `otlp_http_span_exporter.dart:291-298`.

## How to correct it

1. Add 502 and 504 to the list in all three exporters.
2. Carry the numeric status code on a custom exception type. Read that field instead of the message text.

## Notes

- Found by: Fable 5 (medium run)
- Source finding files: `fable-medium/otlp-http-retryable-codes-incomplete.md`
- Parent issue: #95

LaPyaeOo:
#### Description

Fix OTLP/HTTP exporters not retrying HTTP 502 and 504 responses.

Previously, retry detection relied on parsing the exception message and only recognized HTTP 429 and 503. This change introduces a shared `OtlpHttpException` that carries the numeric HTTP status code, allowing retry decisions to use the actual HTTP status code.

The retryable HTTP status codes are now 429, 502, 503, and 504 for all three OTLP/HTTP exporters:

* Trace
* Metrics
* Logs

#### Link to tracking issue

Fixes #224

#### Testing

Added retry tests for HTTP 502 and 504 for all three OTLP/HTTP exporters.

The tests verify that:

* HTTP 502 is retried and succeeds on the second attempt.
* HTTP 504 is retried and succeeds on the second attempt.
* Existing retry behavior continues to pass for HTTP 429 and 503.
* Non-retryable HTTP errors continue to follow the existing behavior.

All OTLP/HTTP retry tests pass: **60/60**.

#### CHANGELOG

* [x] `CHANGELOG.md` updated, or not needed for this change

#### Documentation

Not applicable. No documentation changes are required.

#### Spec Compliance

This change implements the retryable HTTP status codes required by the OpenTelemetry Protocol specification for transient HTTP errors:

https://opentelemetry.io/docs/specs/otlp/#otlphttp-response

The exporter now treats HTTP 429, 502, 503, and 504 as retryable responses.

#### Local Verification

* [x] `dart analyze` passes with no issues.
* [x] OTLP/HTTP retry tests pass: 60/60.
* [x] `git diff --check` passes.

#### Authorship

* [ ] I, a human, authored this pull request and stand behind these changes.
* Assisted-by (AI tools used such as "Fable 5", or "None"): ChatGPT


lib/src/export/otlp_http_exception.dart
```diff
@@ -0,0 +1,8 @@
+import 'package:http/http.dart' as http;
+
+class OtlpHttpException extends http.ClientException {
+  final int statusCode;
+
+  OtlpHttpException(String message, this.statusCode, [Uri? uri])
+      : super(message, uri);
```
yuzurihaaa: iirc, you can use 

`OtlpHttpException(super.message, this.statusCode, [super.uri]);` for this.

yuzurihaaa (changes requested):
The changes in this PR deserves a changelog. Let's add into the CHANGELOG.

